PR #12, opened 2022-06-29
8 build code with strict flags

pbartholomew08:


rfj82982:
I check on Mac and openmpi works with `BUILD=dev` only issue is still mpich. I cannot test yet nvhpc needs to wait for SCARF to be back.

pbartholomew08:
OK, I will leave with you - I’ve done a quick “check it compiles with nvhpc and BUILD=debug/dev” but nothing more.

mathrack:
Here is a suggestion for a patch allowing compilation with MPICH

[git_diff_patch.txt](https://github.com/xcompact3d/2decomp-fft/files/9286719/git_diff_patch.txt)

pbartholomew08:
@mathrack's patch compiles here on ARCHER2 (gnu10 + MPICH).

My only concern would be that I don't think that the code gnu10+ with MPICH is vomiting on is wrong... The alternative would just be to weaken the BUILD=dev FFLAGS for that combination and suggest using OpenMPI for development work and testing with MPICH

mathrack:
If the compiler flags are changing with the MPI library (in debug / dev builds), what about intel MPI ?

pbartholomew08:
> If the compiler flags are changing with the MPI library (in debug / dev builds), what about intel MPI ?

Intel (I think most vendors MPI libraries) are based on MPICH - the PR has different debug/dev flags per compiler for

Intel (MPICH)
Cray (MPICH)
GNU (OpenMPI & MPICH)
NVHPC (MPICH)

Intel also required some slightly looser flags to allow building in "dev" mode - similarly to GNU+MPICH as far as I can tell

mathrack:
Maybe we should clarify in the readme or somewhere else the current supported options / combinations for compilers / MPI libraries. In my opinion, one could use the intel Fortran compiler with Intel MPI / MPICH / OpenMPI.

pbartholomew08:
I agree. In another comment I've suggested that we suggest development is
done with openmpi I just haven't got round to updating the readme

p-costa:
Hi all,

Perhaps appropriate here, while you are at it...

I needed a space here: https://github.com/xcompact3d/2decomp-fft/blob/main/Makefile#L75

```diff
-FFLAGS += $(MODFLAG)$(DECOMPINC) -I$(DECOMPINC)
+FFLAGS += $(MODFLAG) $(DECOMPINC) -I$(DECOMPINC)
```
to be able to build the code with the intel compiler, else I get this warning:

```
mpiifort  -fpp -O3 -mavx2 -march=core-avx2 -mtune=core-avx2 -modulemod -Imod  -DVERSION=\"07234a7\" -DDOUBLE_PREC -cpp  -c src/decomp_2d.f90 -o obj/decomp_2d.o
ifort: command line warning #10006: ignoring unknown option '-modulemod'
```

rfj82982:
Hi Guys I have tested compilations/build with nvhpc with dev/multicore/gpu and it is building. I have included also the `-D_GPU` among the default flags for GPU compilation (i.e. cudaMPI). With the only exception of dev mpich all should build. 
Personal opinion is that we should approve and merge the code. Is this OK for everyone?

rfj82982:
Just a suggestion at the moment we are building for cc80 (i.e. A100) only we should in theory add a variable to choose a different GPU if necessary. Shall we do this in this commit or we do a dedicated commit?

rfj82982 (approved):
Ok we are good to merge

PR #13, opened 2022-07-01
Separate compiler definitions from main Makefile

pbartholomew08:
This will

1) Make the main Makefile easier to read
2) Can be optionally include (e.g. if compilers are already defined,
   don't redefine them)


mathrack (approved):
OpenMP flags should be removed later on

PR #22, opened 2022-08-01
Add MODFLAGS flag

pbartholomew08:
Intel compilers don't support -J (controlling location of generated modules) - to my knowledge all others do

Expected behaviour: all generated *.mod, *.smod files should be located under `mod/`.


mathrack (dismissed):

pbartholomew08:
Use of `-J/-module` (above) places modules/submodules in a `mod/` directory, still need to move source to `src/`

mathrack (approved):

pbartholomew08:
Note, unknown whether flag is correct for Nvidia compilers

rfj82982:
Also nvhpc is complaining about -Jmod

rfj82982:
I have also some other issues. I'll fork from this branch (i was using 8-build-code) and sort issues with compilers, but only for CPU.

pbartholomew08:
Because it wasn't there, I wasn't sure if ``-J`` would be compatible (I should have RTFM!)

rfj82982:
HI Paul I have issue in compiling just using gcc12 and both mpich4.0.1 and openmpi4.1

rfj82982:
`[rfj82982@localhost:2decomp-fft$]mpif90 -v 
Using built-in specs.
COLLECT_GCC=/usr/bin/gfortran
COLLECT_LTO_WRAPPER=/usr/lib64/gcc/x86_64-suse-linux/12/lto-wrapper
OFFLOAD_TARGET_NAMES=nvptx-none:amdgcn-amdhsa
OFFLOAD_TARGET_DEFAULT=1
Target: x86_64-suse-linux
Configured with: ../configure --prefix=/usr --infodir=/usr/share/info --mandir=/usr/share/man --libdir=/usr/lib64 --libexecdir=/usr/lib64 --enable-languages=c,c++,objc,fortran,obj-c++,ada,go,d,jit --enable-offload-targets=nvptx-none,amdgcn-amdhsa, --enable-offload-defaulted --without-cuda-driver --enable-host-shared --enable-checking=release --disable-werror --with-gxx-include-dir=/usr/include/c++/12 --enable-ssp --disable-libssp --disable-libvtv --enable-cet=auto --disable-libcc1 --enable-plugin --with-bugurl=https://bugs.opensuse.org/ --with-pkgversion='SUSE Linux' --with-slibdir=/lib64 --with-system-zlib --enable-libstdcxx-allocator=new --disable-libstdcxx-pch --enable-libphobos --enable-version-specific-runtime-libs --with-gcc-major-version-only --enable-linker-build-id --enable-linux-futex --enable-gnu-indirect-function --program-suffix=-12 --without-system-libunwind --enable-multilib --with-arch-32=x86-64 --with-tune=generic --with-build-config=bootstrap-lto-lean --enable-link-mutex --build=x86_64-suse-linux --host=x86_64-suse-linux
Thread model: posix
Supported LTO compression algorithms: zlib zstd
gcc version 12.1.0 (SUSE Linux) 
[rfj82982@localhost:2decomp-fft$]make 
mpif90  -cpp -fallow-argument-mismatch -O3 -march=native -Jmod -Imod  -DDOUBLE_PREC -DVERSION=\"448968f\"  -c ./decomp_2d.f90 -o obj/./decomp_2d.o
mpif90  -cpp -fallow-argument-mismatch -O3 -march=native -Jmod -Imod  -DDOUBLE_PREC -DVERSION=\"448968f\"  -c ./d2d_log.f90 -o obj/./d2d_log.o
mpif90  -cpp -fallow-argument-mismatch -O3 -march=native -Jmod -Imod  -DDOUBLE_PREC -DVERSION=\"448968f\"  -c ./glassman.f90 -o obj/./glassman.o
./glassman.f90:17:7:

   17 |   use decomp_2d, only : mytype
      |       1
Fatal Error: File ‘decomp_2d.mod’ opened at (1) is not a GNU Fortran module file
compilation terminated.
make: *** [Makefile:87: obj/./glassman.o] Error 1`

pbartholomew08:
...
> 
> 17 | use decomp_2d, only : mytype | 1 Fatal Error: File ‘decomp_2d.mod’ opened at (1) is not a GNU Fortran module file compilation terminated. make: *** [Makefile:87: obj/./glassman.o] Error 1`

That sounds like a module file is not being cleaned up, 1 sec.

rfj82982:
I did a make clean before and i have also manually removed the mod and obj dirs

pbartholomew08:
Can you do ``make clean`` followed by ``ls *.mod mod/*.mod`` to ensure they're all removed properly before rebuilding?

rfj82982:
Sorry I works. but some of the .mod in the main directory did remain. Don't know why. I didn't notice them before and didn't notice they were not removed with make clean

pbartholomew08:
> Sorry I works. but some of the .mod in the main directory did remain. Don't know why. I didn't notice them before and didn't notice they were not removed with make clean

I think they would be left over from an old (i.e. not this branch) build as it won't clean them, 1 sec and I will clean that up

pbartholomew08:
Conflicts should be resolved now

pbartholomew08:
Yes, idea is top level directory should remain fairly "clean" - if happy that it works we can merge into ``main`` and work from there? I don't think branching off branches is the best idea as liable to get compex which is the "true" branch

rfj82982:
nvhpc does not like 
`write (io_unit, '(" Version of the MPI library : ",I0,".",I0)') MPI_VERSION, MPI_SUBVERSION`

`NVFORTRAN-S-0038-Symbol, mpi_version, has not been explicitly declared (src/d2d_log.f90)
NVFORTRAN-S-0038-Symbol, mpi_subversion, has not been explicitly declared (src/d2d_log.f90)`

From documentation we could also get this using 
`MPI_Get_version(version, subversion, ierror) `

rfj82982:
What do you think?

pbartholomew08:
Your suggestion sounds good to me - ``MPI_(SUB)VERSION`` might be implementation-specific macros, MPICH and OpenMPI docs both include ``MPI_Get_version`` you suggest

rfj82982:
Ok tested your version fix/intel-options and works with nvhpc with both CPU only and multicore. I'll not look at the moment to GPU

rfj82982:
Ok works as it is compiling I still need to run the tests

rfj82982:
I have issues in running the examples. With very weird errors. Question is the make clean suppose to remove libdecomp2d.a?

pbartholomew08:
> I have issues in running the examples. With very weird errors. Question is the make clean suppose to remove libdecomp2d.a?

``make clean`` should remove libdecomp2d.a - the examples at the moment are build-only, not running yet (they were added on the main branch) and this branch isn't intended to fix that

pbartholomew08:
I've opened an issue #28 to address getting the examples to run.

rfj82982:
Not sure since I think make clean looks for libdecmp2d withut the .a
`rm -f obj/*.o mod/*.mod mod/*.smod decomp2d
rm -f ./*.o ./*.mod ./*.smod # Ensure old files are removed`

pbartholomew08:
Good point - fixed in above

rfj82982:
Ok I'll have a look on building the example tomorrow/friday. Tomorrow I have to go to Uni York. The make example is doing very weird stuffs like trying to all files that should not. Have you tested it with fftw instead of generic? I was trying to use that together with nvhpc

pbartholomew08:
No - but that would be an issue for another branch

rfj82982:
Ok I manage to get the test exe done using generic fft. I cannot run the test for both c2c and r2c it should be simply
mpirun -n 4 ./fft_test_r2c correct?

pbartholomew08:
They should be but I've not yet run them on the main branch - there was a need to get the code added in it might still need fixing up

rfj82982:
Sorry for being so annoying. The make clean is not cleaning the examples.

rfj82982:
I mean the make clean from root

rfj82982:
Ok even with gnu and openmpi i cannot run the r2c I'll fix another time. At least with nvhpc we are at the same point.

rfj82982 (approved):
I think now it is ok. it compiles also with nvhpc. We'll fix the example issues in the dedicated pull request.

PR #30, opened 2022-08-09
Print environment variables in debug mode

mathrack:
Only rank 0 will print the environment variables.

### DIFF
--- a/src/d2d_log.f90
+++ b/src/d2d_log.f90
@@ -30,8 +30,7 @@ submodule (decomp_2d) d2d_log
 
     ! Local variable
     integer :: io_unit
-
-    integer :: version, subversion, ierr
+    integer :: version, subversion, ierror
 
     !
     ! Default : only rank 0 will print a listing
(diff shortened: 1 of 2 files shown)


src/d2d_log.f90
```diff
@@ -131,6 +131,28 @@ module subroutine d2d_listing(given_io_unit)
...
+    ! The system call, if writing to a file, is not blocking if supported
+    !
+    if (nrank == 0) then
+       write (io_unit, *) '============== Environment variables ======================'
+       write (io_unit, *) '==========================================================='
+       write (io_unit, *) '==========================================================='
+       if (io_unit == output_unit ) then
+          call execute_command_line("env", wait = .true.)
+       else
+          inquire(unit = io_unit, name = fname, iostat = ierror)
```
pbartholomew08, line 147: Probably a dumb question - does this set fname? Otherwise it appears unset

mathrack: It should set fname. See the link below. This might not be bulletproof but it was working in my test.

[https://www.intel.com/content/www/us/en/develop/documentation/fortran-compiler-oneapi-dev-guide-and-reference/top/language-reference/file-operation-i-o-statements/inquire-statement-specifiers/inquire-name-specifier.html#inquire-name-specifier](https://www.intel.com/content/www/us/en/develop/documentation/fortran-compiler-oneapi-dev-guide-and-reference/top/language-reference/file-operation-i-o-statements/inquire-statement-specifiers/inquire-name-specifier.html#inquire-name-specifier)

mathrack: Marking the conversation as resolved. The option to add the name of the file as an argument remains open.

pbartholomew08:
I've suggested a minor update to this on a PR to mathrack/2decomp-fft:main

mathrack:
I have merged your PR. But I am not sure I want to create O(1M) files when running a case !

pbartholomew08:
Probably not, but from experience things stopping working at 999 is annoying. O(1M) is the most I'm aware of Xcompact3d having run up to

pbartholomew08 (approved):

PR #39, opened 2022-08-16
Automatic compilation of the library

mathrack:


mathrack:
As a side remark, it would be interesting to also check automatically the following builds :

- mas-os build
- fftw build

mathrack:
@rfj82982 any modification needed ? In addition to compiling the examples, do we want to run them ?

.github/workflows/Build.yml
```diff
@@ -0,0 +1,112 @@
...
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y intel-oneapi-common-vars intel-oneapi-compiler-fortran intel-oneapi-mkl intel-oneapi-mpi intel-oneapi-mpi-devel
+
+    # Runs a set of commands using the runners shell
+    - name: Compile
+      run: source /opt/intel/oneapi/setvars.sh && make CMP=intel BUILD=debug
+
+    - name: Examples
+      run: source /opt/intel/oneapi/setvars.sh && make examples CMP=intel BUILD=debug
```
pbartholomew08, line 112: Does this (and anything similar) need some sort of time-out? I think there's a limit how much resource we can consume, I don't know what happens if you exceed that but probably don't want it spinning on a stuck test

pbartholomew08: Ah ok - LGTM then!

PR #40, opened 2022-08-19
Fix compiler warnings in restore io branch

pbartholomew08:
Mostly this is where a subroutine has an argument that is used by
ADIOS2 but not MPIIO or vice-versa.
Also discovered some unused variables


pbartholomew08:
N.B. unchecked whether builds against ADIOS2

pbartholomew08:
Compiles with/without ADIOS2 both OpenMPI and MPICH with gfortran 9.4

mathrack (approved):

PR #41, opened 2022-09-05
decomp_2d can use a MPI communicator different from MPI_COMM_WORLD

mathrack:
This modification is proposed so that decomp_2d can use a specific given MPI communicator instead of MPI_COMM_WORLD


pbartholomew08 (approved):

pbartholomew08:
LGTM!

PR #43, opened 2022-09-07
Fix examples

mathrack:
* Fix most of the examples. Compilation with gcc tested. The file fftdata is missing in fft_test_r2c

* The library can run even if the external code does not initialize nrank and nproc

* Examples can run with make check

* Update README.md

* Add a test to detect memory leaks

* Fix the example test2d

* Add a "memory" to Makefile execution

This allows building the library, then later testing without needing to specify the options again

* Update README to explain Makefile.settings

* Minor update for the fft_c2c example

* Rewrite halo test to run in memory (without IO)

Also change prow,pcol to 0 so it runs on arbitrary core count

* Abort halo test with non-zero error on failure

* Use decomp_2d_abort instead of MPI_abort in case of error

* Update Makefile for example/halo_test

* example/halo_test : limited output in the build is not debug

* Make tests more flexible wrt parallelism

Use prow=pcol=0

Define MPIRUN, NP (default mpirun, 1) to set mpirun program and number of ranks to run tests

* decomp_2d_comm is needed inside decomp_2d_abort and decomp_2d_warning

Co-authored-by: Paul Bartholomew <p.bartholomew@epcc.ed.ac.uk>
Co-authored-by: Paul <ptb0890@gmail.com>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,8 @@
 
 # Cmake
 build
+
+# Log files
+*.log
+
+Makefile.settings
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,7 @@ FCFLAGS ?= # user can set default compiler flags
 LDFLAGS ?= # user can set default linker flags
 FFLAGS = $(FCFLAGS)
 LFLAGS = $(LDFLAGS)
-MODFLAG = -J
+MODFLAG = -J 
 
 LIBDECOMP = decomp2d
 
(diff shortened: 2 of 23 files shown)


mathrack:
Running the test fft_c2c on one CPU, the error is exactly zero. On more than one CPU, the error is 1e-8 in case of double precision.

rfj82982 (approved):
Tests are now ok

PR #50, opened 2022-09-21
Build test cases under strictrer (BUILD=debug/dev) flags

pbartholomew08:
This prevents test building failing when stricter debug flags are enabled

### DIFF
--- a/examples/fft_test_c2c/Makefile
+++ b/examples/fft_test_c2c/Makefile
@@ -10,8 +10,11 @@ OBJ = fft_test_c2c.o
 MPIRUN ?= mpirun
 NP ?= 1
 
-fft_test_c2c: $(OBJ)
+fft_test_c2c: $(DECOMPINC) $(OBJ)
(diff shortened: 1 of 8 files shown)


mathrack:
Currently, the test cases do not contain case-specific modules or submodules. Thus, there is no need for creating a folder, it is going to stay empty...

mathrack:
I suggest the following fix : 
[patch.txt](https://github.com/xcompact3d/2decomp-fft/files/9622721/patch.txt)

mathrack:
I will try that but it would be easier to merge the automatic compilation to check proper compilation with gnu / intel  / nvidia

pbartholomew08:
I think if `$(MODFLAG)` works with GNU we can assume it works with others and fix later if not

pbartholomew08:
Or we make an empty directory ;)

pbartholomew08:
@mathrack this should be ready now - it builds examples under `BUILD=dev` with GNU and Intel compilers

mathrack (approved):

PR #53, opened 2022-09-22
Update profiler_caliper.f90

mathrack:
Remove unused variables to allow `dev` builds

### DIFF
--- a/src/profiler_caliper.f90
+++ b/src/profiler_caliper.f90
@@ -65,6 +65,10 @@ submodule (decomp_2d) d2d_profiler_caliper
    !
    ! The setup of the profiler can be modified before init
    !
+   ! TODO Current setup is limited to "runtime-report"
+   !      Advance setup is possible using
+   !         - extra optional argument of type char at prepare stage
+   !         - env. variable CALI_CONFIG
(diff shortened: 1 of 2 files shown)


src/profiler_caliper.f90
```diff
@@ -127,9 +124,9 @@ subroutine manager_error(mmm)
       logical :: ret
       character(len=:), allocatable :: errmsg
 
-      ret = mgr%error()
+      ret = mmm%error()
```
pbartholomew08, line 127: Why not call `mmm` -> `mgr`? It would be easier to read I think.

pbartholomew08 (changes requested):

src/profiler_caliper.f90
pbartholomew08: Or if concerned `mgr` might clash with the module-level variable, another name than `mmm` would be prefereable

mathrack: Sounds good, I will use `manager`

mathrack: Fixed

pbartholomew08 (approved):
LGTM!

PR #54, opened 2022-09-22
Automatically add -DPROFILER

pbartholomew08:


mathrack (changes requested):
It looks good for merging. Please update the line 
``PROFILER = none# none, caliper (add -DPROFILER to DEFS if not none)``
to remove the comment about -DPROFILER. Please also update the README to remove the required update of the variables DEFS.

Makefile
```diff
@@ -61,6 +61,10 @@ else ifeq ($(FFT),cufft)
 endif
 
 ### Add the profiler if needed
+ifeq ($(PROFILER),none)
```
mathrack, line 64: ifneq ?

mathrack, line 64: ifneq ?

mathrack (changes requested):

Makefile
pbartholomew08: ah, wasn't sure that existed

mathrack (approved):

PR #56, opened 2022-09-22
48 fix update halo

pbartholomew08:


pbartholomew08:
I've added an `optional` parameter to `update_halo` indicating which pencil you are in. This avoids breaking changes to existing code, however as the test shows existing code may be broken, perhaps a breaking change is to be preferred.

mathrack:
Do we need tests on `ipencil` there also ?

https://github.com/xcompact3d/2decomp-fft/blob/48-fix-update_halo/src/halo_common.f90#L150

Breaking compatibility should be discussed with more people. I suggest we remain compatible and simply add, where relevant, `call decomp_2d_warning(0, "This version of the halo subroutine is no longer supported, see commit xxx")` ?

mathrack (changes requested):
See my last comment on the discussion tab

pbartholomew08:
> Do we need tests on `ipencil` there also ?
> 
> https://github.com/xcompact3d/2decomp-fft/blob/48-fix-update_halo/src/halo_common.f90#L150

Ah yes, I somehow missed those ones, thought it also implies the tests don't catch that

PR #59, opened 2022-09-24
Add type-bound procedures in decomp_info

mathrack:


pbartholomew08:
This is quite a significant change to the semantics - I think we need to discuss at the hackathon

mathrack:
Please note that the original subroutines `decomp_info_init` and `decomp_info_finalize` are still present and usable.

pbartholomew08:
I think one way would be better than two

mathrack:
I have kept them to avoid breaking existing codes.

pbartholomew08:
I think it would be best not to have a mishmash of codes - let's see what others think about the different approaches and choose one.

mathrack:
Any update needed on this PR ?

mathrack:
It looks like there is no interest for this PR, I will close it.

PR #61, opened 2022-09-27
Add fftw3 to the automatic build process

mathrack:
Also fix missing variable FC in Makefile.settings.

It would be nice to add the profiler caliper to the automatic build process, but it seems to lack a Debian / Ubuntu package

### DIFF
--- a/.github/workflows/Build.yml
+++ b/.github/workflows/Build.yml
@@ -12,7 +12,7 @@ on:
 
 # A workflow run is made up of one or more jobs that can run sequentially or in parallel
 jobs:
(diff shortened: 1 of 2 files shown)


.github/workflows/Build.yml
```diff
@@ -33,7 +33,7 @@ jobs:
       run: make BUILD=debug
 
     - name: Examples
-      run: make examples BUILD=debug
+      run: make examples
```
pbartholomew08, line 36: Why remove `BUILD=debug`? I think this is more useful here as it will catch my sloppy code ;)

pbartholomew08: Oh, wait I see this is only for the examples, please ignore

pbartholomew08 (approved):
LGTM

PR #72, opened 2022-10-12
Add Intel MKL FFT Backend

mathrack:
Please note that the intel CI does not yet support the MKL because the file `mkl_dfti.f90` can not be located. This file should be provided with the intel MKL library. See [https://github.com/geodynamics/relax/blob/master/src/mkl_dfti.f90](https://github.com/geodynamics/relax/blob/master/src/mkl_dfti.f90) for instance

EDIT : please squash and merge to erase the commit history

### DIFF
--- a/.github/workflows/Build_all.yml
+++ b/.github/workflows/Build_all.yml
@@ -102,7 +102,7 @@ jobs:
 
     # Runs a set of commands using the runners shell
     - name: Compile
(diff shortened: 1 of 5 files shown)


mathrack:
The PR is ready for review but needs to be tested.

mathrack:
As a side comment, it looks like the MKL FFT engine does not support the OVERWRITE flag for inplace FFT.

pbartholomew08 (approved):
Would prefer just pointing Makefile at mkl_dfti source, but this works too

PR #73, opened 2022-10-13
Don't pull MKL files into 2decomp source

pbartholomew08:


mathrack (approved):
Looks ok for merging. Do you think this will work for CI ?

pbartholomew08:
Not sure, hopefully tidier. But something seems wrong with Nvidia currently

PR #80, opened 2022-10-19
Delete old mem_leak target referencing timing in fft_physical_z

pbartholomew08:
I'm assuming this is a clear-cut old piece of old code but thought worth checking with you @rfj82982


rfj82982:
Yes this should have been fixed I’ll merge it now.

rfj82982 (approved):

PR #81, opened 2022-10-19
Fix fftw3 CI  build

mathrack:


p-costa:
Hello,

It may be better to use `type(C_PTR)` instead:
https://fftw.org/doc/FFTW-Fortran-type-reference.html

mathrack:
Thank you @p-costa, `type(c_ptr)` is used in the file fftw_f03.f90, this is probably the best solution

pbartholomew08 (approved):
LGTM!

PR #92, opened 2022-11-09
Check cuda error code nvhpc ok

rfj82982:
Here the new pull request for the cuda error code. The deposit for the submodule is in the deposit branch


mathrack (approved):

rfj82982:
I am still resolving the conflicts. I’ll ask to review them once finish.

rfj82982:
Hi Guys I solved all the conflicts and now it is working. Testing done: 

1. GCC12 + mpich OK
2. GCC12 + mpich+FFTW3 OK
3. NVHPC CPU OK
4. NVHPC GPU OK
5. NVHPC GPU NCCL OK (up to 4 GPUs) OK
6. Intel (but without mkl) OK
7. GCC 11.3 + OMPI 4.1.4 OK

Only test not OK is 

1.  GCC 12 + OMPI 4.1.2 

Can someone check again with open MPI since the error is very weird

examples/halo_test/halo_test.f90
```diff
@@ -86,7 +86,7 @@ subroutine initialise()
 
     call random_seed(size = n)
     allocate(seed(n))
-    seed = nrank
+    seed = nrank+1
```
pbartholomew08, line 89: Why this change? Just thinking it could change people's results (slightly)

src/decomp_2d_init_fin.f90
```diff
@@ -27,8 +19,9 @@
   !     all internal data structures initialised properly
   !     library ready to use
   !======================================================================
-  module subroutine decomp_2d_init_ref(nx,ny,nz,p_row,p_col,periodic_bc,comm)
+  subroutine decomp_2d_init_ref(nx,ny,nz,p_row,p_col,periodic_bc,comm)
 
+    use mpi
```
pbartholomew08, line 24: Now that this file is ``#included`` is this still necessary? Sometimes Cray IIRC won't compile if a module is ``used`` twice

pbartholomew08:
Looks good to me, only 2 minor comments that may be non-issues.

examples/halo_test/halo_test.f90
rfj82982: This is for NVHPC that does not like seed starting from 0. I modified this also in Xcompact3d a while ago. In terms of results should be ok since compute the divergence of an initial field generated by rand. The initial field is the same for both methods just computed in two different ways (transpose and halo) therefore the output must be the same.

pbartholomew08: ah - I thought this was being set elsewhere, not in the test case!

pbartholomew08 (approved):
LGTM!

PR #96, opened 2022-11-16
Fix to avoid bug when the null() pointer is provided to d…

mathrack:
…ecomp_info_init

### DIFF
--- a/src/fft_common.f90
+++ b/src/fft_common.f90
@@ -91,6 +91,7 @@ implicit none
 integer, intent(IN) :: pencil
 integer, intent(IN) :: nx, ny, nz
 
+type(decomp_info), target :: tmp
 integer :: status, errorcode
 
 #ifdef PROFILER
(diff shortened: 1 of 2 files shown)


mathrack:
This PR is ready for review

src/fft_common.f90
```diff
@@ -120,6 +120,7 @@ subroutine fft_init_general(pencil, nx, ny, nz)
 if (nx_fft==nx_global.and.ny_fft==ny_global.and.nz_fft==nz_global) then
 ph=>decomp_main
 else
+if (.not.associated(ph)) allocate(ph)
```
pbartholomew08, line 123: I don't understand this - what does ph point at after this?

pbartholomew08: after reading/asking around sets aside space for object

pbartholomew08 (approved):

PR #102, opened 2022-11-29
More options for the log output

### LINKED ISSUE #100: make creation of `decomp_2d_setup.log` file optional?
I think many users (myself included) like all the outputs to be lumped into a single log file by redirecting stdout to a file, so I wonder if there could/should be a way to have the output `decomp_2d_setup.log` in stdout? (Perhaps a less verbose option? Though the current level of verbosity seems quite reasonable.)

**Describe a possible solution**
I think something like this would work:

```diff
diff --git a/src/decomp_2d_init_fin.f90 b/src/decomp_2d_init_fin.f90
index fda17fd..d43e1aa 100644
--- a/src/decomp_2d_init_fin.f90
+++ b/src/decomp_2d_init_fin.f90
@@ -213,12 +213,8 @@
     write(fname, "(I0)") nrank ! Adapt to magnitude of nrank
     open(newunit=iounit, file='decomp_2d_setup_'//trim(fname)//'.log', iostat=ierror)
 #else
-    if (nrank == 0) then
-       open(newunit=iounit, file="decomp_2d_setup.log", iostat=ierror)
-    else
-       iounit = output_unit
-       ierror = 0
-    endif
+    iounit = output_unit
+    ierror = 0
 #endif
```
The default output of `decomp_2d_setup.log` would be in stdout, unless in debugging mode, where a filename per task is written. 

Not sure if this is a change you'd like, perhaps it is more a matter of what one is used to. I don't mind just changing my fork instead.

Edit: oh, wow, issue number 100 :tada:

mathrack:
closes #100

### DIFF
--- a/src/decomp_2d.f90
+++ b/src/decomp_2d.f90
@@ -84,16 +84,14 @@ module decomp_2d
   !
   ! The default value is 2 (3 for debug builds)
   !
-  enum, bind(c)
-     enumerator :: D2D_LOG_NO = 0
-     enumerator :: D2D_LOG_STDOUT = 1
-     enumerator :: D2D_LOG_FILE = 2
-     enumerator :: D2D_LOG_ALL = 3
-  end enum
+  integer, parameter, public :: D2D_LOG_NO = 0
+  integer, parameter, public :: D2D_LOG_STDOUT = 1
+  integer, parameter, public :: D2D_LOG_FILE = 2
+  integer, parameter, public :: D2D_LOG_ALL = 3
 #ifdef DEBUG
-  integer(kind(D2D_LOG_NO)), public, save :: decomp_log = D2D_LOG_ALL
+  integer, public, save :: decomp_log = D2D_LOG_ALL
 #else
-  integer(kind(D2D_LOG_NO)), public, save :: decomp_log = D2D_LOG_FILE
+  integer, public, save :: decomp_log = D2D_LOG_FILE
 #endif
 
   !
--- a/src/decomp_2d.f90
+++ b/src/decomp_2d.f90
@@ -84,14 +84,14 @@ module decomp_2d
   !
   ! The default value is 2 (3 for debug builds)
   !
-  integer, parameter, public :: D2D_LOG_NO = 0
+  integer, parameter, public :: D2D_LOG_QUIET = 0
   integer, parameter, public :: D2D_LOG_STDOUT = 1
-  integer, parameter, public :: D2D_LOG_FILE = 2
-  integer, parameter, public :: D2D_LOG_ALL = 3
+  integer, parameter, public :: D2D_LOG_TOFILE = 2
+  integer, parameter, public :: D2D_LOG_TOFILE_FULL = 3
 #ifdef DEBUG
-  integer, public, save :: decomp_log = D2D_LOG_ALL
+  integer, public, save :: decomp_log = D2D_LOG_TOFILE_FULL
 #else
-  integer, public, save :: decomp_log = D2D_LOG_FILE
+  integer, public, save :: decomp_log = D2D_LOG_TOFILE
 #endif
 
   !
--- a/src/decomp_2d.f90
+++ b/src/decomp_2d.f90
@@ -75,6 +75,28 @@ module decomp_2d
   logical, save :: periodic_x, periodic_y, periodic_z
 
   !
+  ! Output for the log can be changed by the external code before calling decomp_2d_init
+  !
+  !    0 => No log output
+  !    1 => Master rank log output to stdout
+  !    2 => Master rank log output to the file "decomp_2d_setup.log"
+  !    3 => All ranks log output to a dedicated file
+  !
+  ! The default value is 2 (3 for debug builds)
+  !
+  enum, bind(c)
+     enumerator :: D2D_LOG_NO = 0
+     enumerator :: D2D_LOG_STDOUT = 1
+     enumerator :: D2D_LOG_FILE = 2
+     enumerator :: D2D_LOG_ALL = 3
+  end enum
+#ifdef DEBUG
+  integer(kind(D2D_LOG_NO)), public, save :: decomp_log = D2D_LOG_ALL
+#else
+  integer(kind(D2D_LOG_NO)), public, save :: decomp_log = D2D_LOG_FILE
+#endif
+
+  !
   ! Debug level can be changed by the external code before calling decomp_2d_init
   !
   ! The environment variable "DECOMP_2D_DEBUG" can be used to change the debug level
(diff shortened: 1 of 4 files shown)


src/decomp_2d.f90
```diff
@@ -74,6 +74,28 @@ module decomp_2d
...
+  !    3 => All ranks log output to a dedicated file
+  !
+  ! The default value is 2 (3 for debug builds)
+  !
+  enum, bind(c)
+     enumerator :: D2D_LOG_NO = 0
+     enumerator :: D2D_LOG_STDOUT = 1
+     enumerator :: D2D_LOG_FILE = 2
+     enumerator :: D2D_LOG_ALL = 3
+  end enum
```
pbartholomew08, line 92: Can the user pass these values, e.g. `D2D_LOG_NO` rather than `0` to `decomp_2d_init`? I think this would be better if possible and should be the recommendation in the README too if so.

pbartholomew08 (changes requested):

src/decomp_2d.f90
mathrack: This is not possible with the enum structure. But the code could export a list of parameters, such as the ones above yes. Similarly to what FFTW does, it is probably even possible to generate this list of parameters automatically.

mathrack: The enum was removed in 333ad0b714e361fb57d141b37ae84d4b435b3c64. The code is now easier to call from Fortran and harder to call from C.

p-costa:
Thanks for the PR following my issue!

I thought of the following alternative names, but I am not sure what you think: `NO -> QUITET`; `FILE -> TOFILE`; `ALL` -> `TOFILE_FULL`.

But the current ones also make sense to me, perhaps with an `! inline comment documenting them`

mathrack:
@p-costa feel free to modify this branch or open a PR to change it if needed. As far as I am concerned, no problem for changing the labels.

src/log.f90
```diff
@@ -35,14 +35,10 @@ module subroutine d2d_listing(given_io_unit)
...
-    !
-    ! In DEBUG mode, all ranks will print a listing
-    !
-#ifndef DEBUG
-    if (nrank /= 0) return
-#endif
+    ! Output log if needed
+    if (decomp_log == D2D_LOG_NO) return
+    if (decomp_log == D2D_LOG_STDOUT .and. nrank /= 0) return
+    if (decomp_log == D2D_LOG_FILE .and. nrank /= 0) return
```
p-costa, line 41:
```suggestion
    if (decomp_log == D2D_LOG_TOFILE .and. nrank /= 0) return
```

src/decomp_2d.f90
```diff
@@ -74,6 +74,26 @@ module decomp_2d
...
+  ! Output for the log can be changed by the external code before calling decomp_2d_init
+  !
+  !    0 => No log output
+  !    1 => Master rank log output to stdout
+  !    2 => Master rank log output to the file "decomp_2d_setup.log"
+  !    3 => All ranks log output to a dedicated file
+  !
+  ! The default value is 2 (3 for debug builds)
+  !
+  integer, parameter, public :: D2D_LOG_NO = 0
```
p-costa, line 87:
```suggestion
  integer, parameter, public :: D2D_LOG_QUIET = 0
```

src/decomp_2d.f90
```diff
@@ -74,6 +74,26 @@ module decomp_2d
...
+  integer, parameter, public :: D2D_LOG_STDOUT = 1
+  integer, parameter, public :: D2D_LOG_FILE = 2
```
p-costa, line 89:
```suggestion
  integer, parameter, public :: D2D_LOG_TOFILE = 2
```

src/decomp_2d.f90
```diff
@@ -74,6 +74,26 @@ module decomp_2d
...
+  integer, parameter, public :: D2D_LOG_ALL = 3
```
p-costa, line 90:
```suggestion
  integer, parameter, public :: D2D_LOG_TOFILE_FULL = 3
```

src/decomp_2d.f90
```diff
@@ -74,6 +74,26 @@ module decomp_2d
...
+#ifdef DEBUG
+  integer, public, save :: decomp_log = D2D_LOG_ALL
```
p-costa, line 92:
```suggestion
  integer, public, save :: decomp_log = D2D_LOG_TOFILE_FULL
```

src/decomp_2d.f90
```diff
@@ -74,6 +74,26 @@ module decomp_2d
...
+#else
+  integer, public, save :: decomp_log = D2D_LOG_FILE
```
p-costa, line 94:
```suggestion
  integer, public, save :: decomp_log = D2D_LOG_TOFILE
```

src/decomp_2d_init_fin.f90
```diff
@@ -209,17 +207,15 @@ subroutine decomp_2d_init_ref(nx,ny,nz,p_row,p_col,periodic_bc,comm)
     !
     ! Select the IO unit for decomp_2d setup
     !
-#ifdef DEBUG
-    write(fname, "(I0)") nrank ! Adapt to magnitude of nrank
-    open(newunit=iounit, file='decomp_2d_setup_'//trim(fname)//'.log', iostat=ierror)
-#else
-    if (nrank == 0) then
+    if (decomp_log == D2D_LOG_ALL) then
```
p-costa, line 210:
```suggestion
    if (decomp_log == D2D_LOG_TOFILE_FULL) then
```

src/decomp_2d_init_fin.f90
```diff
@@ -209,17 +207,15 @@ subroutine decomp_2d_init_ref(nx,ny,nz,p_row,p_col,periodic_bc,comm)
...
+       write(fname, "(I0)") nrank ! Adapt to magnitude of nrank
+       open(newunit=iounit, file='decomp_2d_setup_'//trim(fname)//'.log', iostat=ierror)
+    elseif (nrank == 0 .and. decomp_log == D2D_LOG_FILE) then
```
p-costa, line 213:
```suggestion
    elseif (nrank == 0 .and. decomp_log == D2D_LOG_TOFILE) then
```

src/log.f90
```diff
@@ -35,14 +35,10 @@ module subroutine d2d_listing(given_io_unit)
...
-    !
-    ! Default : only rank 0 will print a listing
-    !
-    ! In DEBUG mode, all ranks will print a listing
-    !
-#ifndef DEBUG
-    if (nrank /= 0) return
-#endif
+    ! Output log if needed
+    if (decomp_log == D2D_LOG_NO) return
```
p-costa, line 39:
```suggestion
    if (decomp_log == D2D_LOG_QUIET) return
```

p-costa (changes requested):

pbartholomew08 (approved):

mathrack:
Looks good, merging

PR #105, opened 2022-11-30
Update caliper version

mathrack:


pbartholomew08:
Does v2.8.0 still work? Might be worth stating the version(s) that have been checked in the README, although leave the instructions with the latest/preferred version

mathrack:
I have tested the new version and it works. The previous version is probably still working yes.

pbartholomew08 (approved):

PR #119, opened 2022-12-13
Describe preprocessor variables in the README

### LINKED ISSUE #74: Clarify the validity of Compile flags
Hello,

The 2decomp-fft library is using a relatively large number of compile-time flags. See [here](https://github.com/xcompact3d/2decomp-fft/blob/main/src/d2d_log.f90#L80) for a probably complete list.

Some flags can lead to conflicts, or unexpected behavior. For instance, GPU / NCCL flags are not active when the flag SHM or EVEN are provided.

Also, some flags are not supported by all FFT engines. For instance, MKL backend does not support the OVERWRITE flag.

If possible, this should be clarified or documented.

mathrack:
closes #74

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,7 +113,7 @@ This preprocessor variable is not valid for GPU builds. It leads to padded allto
 
 #### OVERWRITE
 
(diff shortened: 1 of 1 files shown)


pbartholomew08 (approved):
LGTM!

PR #122, opened 2022-12-15
Expose fft backend

### LINKED ISSUE #117: Expose the FFT backend in a public variable
No problem

**Describe the solution you'd like**
The 2decomp library could expose a public variable so the external code is aware of the FFT backend selected at compile time.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

mathrack:
closes #117

### DIFF
--- a/src/decomp_2d_constants.f90
+++ b/src/decomp_2d_constants.f90
@@ -69,6 +69,5 @@ module decomp_2d_constants
    integer, parameter, public :: D2D_FFT_BACKEND_MKL = 3
    integer, parameter, public :: D2D_FFT_BACKEND_CUFFT = 4
 
-
 end module decomp_2d_constants
 
--- a/src/decomp_2d_constants.f90
+++ b/src/decomp_2d_constants.f90
@@ -1,0 +1,74 @@
+!=======================================================================
+! This is part of the 2DECOMP&FFT library
+!
+! 2DECOMP&FFT is a software framework for general-purpose 2D (pencil)
+! decomposition. It also implements a highly scalable distributed
+! three-dimensional Fast Fourier Transform (FFT).
+!
+! Copyright (C) 2009-2012 Ning Li, the Numerical Algorithms Group (NAG)
+! Copyright (C) 2021               the University of Edinburgh (UoE)
+!
+!=======================================================================
+
+! Constants for the 2decomp&fft library
+
+module decomp_2d_constants
+
+   implicit none
+
+   !
+   ! Output for the log can be changed by the external code before calling decomp_2d_init
+   !
+   !    0 => No log output
+   !    1 => Master rank log output to stdout
+   !    2 => Master rank log output to the file "decomp_2d_setup.log"
+   !    3 => All ranks log output to a dedicated file
+   !
+   ! The default value is 2 (3 for debug builds)
+   !
+   integer, parameter, public :: D2D_LOG_QUIET = 0
+   integer, parameter, public :: D2D_LOG_STDOUT = 1
+   integer, parameter, public :: D2D_LOG_TOFILE = 2
+   integer, parameter, public :: D2D_LOG_TOFILE_FULL = 3
+
+   !
+   ! Debug level can be changed by the external code before calling decomp_2d_init
+   !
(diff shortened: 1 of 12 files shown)


mathrack:
This PR needs a few more commits before review

src/fft_cufft.f90
```diff
@@ -36,6 +36,14 @@ module decomp_2d_fft
    integer*4, save :: plan(-1:2, 3)
    complex*8, device, allocatable, dimension(:) :: cufft_workspace
 
+   ! Supported FFT backends
+   integer, parameter, public :: D2D_FFT_BACKEND_GENERIC = 0
+   integer, parameter, public :: D2D_FFT_BACKEND_FFTW3 = 1
+   integer, parameter, public :: D2D_FFT_BACKEND_FFTW3_F03 = 2
+   integer, parameter, public :: D2D_FFT_BACKEND_MKL = 3
+   integer, parameter, public :: D2D_FFT_BACKEND_CUFFT = 4
+   integer, parameter, public :: D2D_FFT_BACKEND = D2D_FFT_BACKEND_CUFFT
```
pbartholomew08, line 45: I think `GENERIC` should be the default value

src/fft_cufft.f90
```diff
@@ -36,6 +36,14 @@ module decomp_2d_fft
    integer*4, save :: plan(-1:2, 3)
    complex*8, device, allocatable, dimension(:) :: cufft_workspace
 
+   ! Supported FFT backends
+   integer, parameter, public :: D2D_FFT_BACKEND_GENERIC = 0
+   integer, parameter, public :: D2D_FFT_BACKEND_FFTW3 = 1
+   integer, parameter, public :: D2D_FFT_BACKEND_FFTW3_F03 = 2
+   integer, parameter, public :: D2D_FFT_BACKEND_MKL = 3
+   integer, parameter, public :: D2D_FFT_BACKEND_CUFFT = 4
```
pbartholomew08, line 44: I would define these once in a separate module, rather than keep redefining

mathrack: The current workaround to avoid repetitions would be to have them defined in decomp_2d. But I do not want to put decomp_2d_fft stuff inside decomp_2d...

mathrack: Or in a file and #include the file

mathrack: Otherwise, we need to add a file with all the public symbols / parameters in a dedicated module ?

pbartholomew08: > Or in a file and #include the file

Definitely not :)

> Otherwise, we need to add a file with all the public symbols / parameters in a dedicated module ?

This is what I would suggest , e.g. `decomp_2d_constants` module. The decomp_2d module is too big, I don't think we should be exposing everything there and should eventually move things into the new module. However as this is a new variable there is no harm in putting it in the right place to begin with

mathrack:
I have tested the generic FFT engine with a dev build

pbartholomew08 (changes requested):

src/decomp_2d_constants.f90
```diff
@@ -0,0 +1,73 @@
...
+!
+! Copyright (C) 2009-2012 Ning Li, the Numerical Algorithms Group (NAG)
+! Copyright (C) 2021               the University of Edinburgh (UoE)
+!
+!=======================================================================
+
+! Constants for the 2decomp&fft library
+
+module decomp_2d_constants
+
```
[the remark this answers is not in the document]
pbartholomew08: My mistake, this is part of f2003

src/fft_generic.f90
```diff
@@ -32,14 +35,23 @@ module decomp_2d_fft
 !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
    subroutine init_fft_engine
 
-      implicit none
+      use iso_fortran_env, only : output_unit
```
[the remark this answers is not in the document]
mathrack: yes, fftw3 is broken currently

mathrack:
I have tested the generic and both fftw3 backends

pbartholomew08 (approved):

PR #127, opened 2022-12-19
Fuse working arrays wk2_c2c and wk2_r2c

mathrack:
The trick could apply to fftw3_f03 as well.

### DIFF
--- a/src/fft_common.f90
+++ b/src/fft_common.f90
@@ -35,8 +35,10 @@ TYPE(DECOMP_INFO), pointer, save :: ph => null()  ! physical space
 TYPE(DECOMP_INFO), target, save :: sp  ! spectral space
 
 ! Workspace to store the intermediate Y-pencil data
-! *** TODO: investigate how to use only one workspace array
-complex(mytype), allocatable, dimension(:, :, :) :: wk2_c2c, wk2_r2c
+complex(mytype), allocatable, target, dimension(:, :, :) :: wk2_c2c
+complex(mytype), pointer, dimension(:,:,:) :: wk2_r2c
+! Workspace for r2c and c2r transforms
+! FIXME could be removed using in-place r2c and c2r ?
 complex(mytype), allocatable, dimension(:, :, :) :: wk13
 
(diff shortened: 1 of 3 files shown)


src/fft_common.f90
```diff
@@ -129,7 +131,7 @@ subroutine fft_init_general(pencil, nx, ny, nz)
    end if
 
    allocate (wk2_c2c(ph%ysz(1), ph%ysz(2), ph%ysz(3)), STAT=status)
-   allocate (wk2_r2c(sp%ysz(1), sp%ysz(2), sp%ysz(3)), STAT=status)
+   call c_f_pointer(c_loc(wk2_c2c), wk2_r2c, sp%ysz)
```
pbartholomew08, line 134: Why make it a C pointer, would ``wk2_rc2 => wk2_c2c`` not work and be clearer?

mathrack: I think this would lead to incorrect bounds on `wk2_r2c`. Using`c_loc` we make sure the shape / bounds is not transferred. The equivalent of the above statement without C pointer is probably

```
wk2_r2c(1:sp%ysz(1), 1:sp%ysz(2), 1:sp%ysz(3)) => wk2_c2c
````

In my opinion the above is clearer.

pbartholomew08: I think the latter is clearer, it is plain Fortran. However, either way I think this needs some explanatory comment about using a different size than what it's pointing at

mathrack: The repetition `1:sp%ysz(1), 1:sp%ysz(2), 1:sp%ysz(3)` is not really my cup of tea. I will add some comments to explain that both array start at the same memory location but have distinct size. Feel free to switch from one form to the other after my commit.

mathrack: See https://github.com/xcompact3d/2decomp-fft/pull/127/commits/62693d3e4bc6f824954b1d61c7d4eac4672088bb

pbartholomew08 (approved):

PR #130, opened 2022-12-26
Upfate fft

mathrack:
The main modification is the addition of a submodule to decom_2d_fft dedicated to the log


mathrack:
This PR is ready for review. I have a sketch allowing in-place r2c and c2r transforms for both fftw3 backends (probably valid for oneMKL as well). I will use a dedicated PR to describe it.

PR #135, opened 2023-01-13
Update fft (part 2)

rfj82982:
I have updated also the tests for 2decomp to adapt for the non memory managed case for GPU. The only missing is the halo since it requires some changes in the routines. 
I’ll also open a feature request to adapt the IO part of 2decomp. In the example of io for single plane only the write for aligned planes works  (i.e. X write for u1). The write plane has some transposes and for the manual memory management we need. This will give at least consistency that when doing IO we move the data to the host and we write from there. 
I would like to do the pull request to allow Wei and I to put properly the sin/cos transform.

### DIFF
--- a/examples/Makefile
+++ b/examples/Makefile
@@ -1,7 +1,7 @@
-.PHONY: init_test test2d fft_physical_x fft_physical_z io_test 
+.PHONY: init_test test2d fft_physical_x fft_physical_z io_test halo_test
 
 # Just build the examples
-examples: init_test test2d fft_physical_x fft_physical_z io_test
+examples: init_test test2d fft_physical_x fft_physical_z io_test halo_test
 	@echo "Built the examples"
 
 init_test:
@@ -14,6 +14,8 @@ fft_physical_z:
 	$(MAKE) -C $@ all
 io_test:
 	$(MAKE) -C $@ all
+halo_test:
+	$(MAKE) -C $@ all
 
 check:
 	cd init_test; $(MAKE) $@
@@ -21,6 +23,7 @@ check:
 	cd fft_physical_x; $(MAKE) $@
 	cd fft_physical_z; $(MAKE) $@
 	cd io_test; $(MAKE) $@
+	cd halo_test; $(MAKE) $@
 
 clean:
 	cd init_test; $(MAKE) $@
@@ -28,5 +31,6 @@ clean:
 	cd fft_physical_x; $(MAKE) $@
 	cd fft_physical_z; $(MAKE) $@
 	cd io_test; $(MAKE) $@
+	cd halo_test; $(MAKE) $@
 
 export
--- a/examples/Makefile
+++ b/examples/Makefile
@@ -1,36 +1,28 @@
-.PHONY: test2d fft_physical_x fft_physical_z halo_test init_test io_test 
+.PHONY: init_test test2d fft_physical_x fft_physical_z 
 
 # Just build the examples
-examples: test2d fft_physical_x fft_physical_z halo_test init_test io_test
+examples: init_test test2d fft_physical_x fft_physical_z 
 	@echo "Built the examples"
 
+init_test:
+	$(MAKE) -C $@ all
 test2d:
 	$(MAKE) -C $@ $@
 fft_physical_x:
 	$(MAKE) -C $@ all
 fft_physical_z:
 	$(MAKE) -C $@ all
-halo_test:
-	$(MAKE) -C $@ $@
-init_test:
-	$(MAKE) -C $@ all
-io_test:
-	$(MAKE) -C $@ all
 
 check:
+	cd init_test; $(MAKE) $@
 	cd test2d; $(MAKE) $@
 	cd fft_physical_x; $(MAKE) $@
 	cd fft_physical_z; $(MAKE) $@
-	cd halo_test; $(MAKE) $@
-	cd init_test; $(MAKE) $@
-	cd io_test; $(MAKE) $@
 
 clean:
+	cd init_test; $(MAKE) $@
 	cd test2d; $(MAKE) $@
 	cd fft_physical_x; $(MAKE) $@
 	cd fft_physical_z; $(MAKE) $@
-	cd halo_test; $(MAKE) $@
-	cd init_test; $(MAKE) $@
-	cd io_test; $(MAKE) $@
 
 export
--- a/examples/Makefile
+++ b/examples/Makefile
@@ -1,7 +1,7 @@
-.PHONY: init_test test2d fft_physical_x fft_physical_z 
+.PHONY: init_test test2d fft_physical_x fft_physical_z io_test 
 
 # Just build the examples
-examples: init_test test2d fft_physical_x fft_physical_z 
+examples: init_test test2d fft_physical_x fft_physical_z io_test
 	@echo "Built the examples"
 
 init_test:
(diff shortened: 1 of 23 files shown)


examples/fft_physical_x/fft_grid_x.f90
```diff
@@ -47,10 +49,15 @@ program fft_physical_x
    ! output is Z-pencil data
    call alloc_x(in, ph, .true.)
    call alloc_z(out, ph, .true.)
+   ! Convert pointers to loops start/end to scalar
```
mathrack, line 52: This is mandatory for GPUs ? This modification is missing in other examples ?

examples/io_test/io_plane_test.f90
```diff
@@ -1,9 +1,15 @@
+!! NB in case of GPU only the writing in the aligned pencil (i.e. X for a 1 array) is performed. 
+!! IO subrotines needs update for non managed GPU case
```
mathrack, line 2: We should write known issues in a dedicated file ? The comments in a example is probably not the best location.

examples/io_test/io_test.f90
```diff
@@ -79,7 +86,7 @@ program io_test
    do k = xstart(3), xend(3)
       do j = xstart(2), xend(2)
          do i = xstart(1), xend(1)
-            if (abs((u1(i, j, k) - u1b(i, j, k))) > eps) stop 1
+            if (abs((u2(i, j, k) - u2b(i, j, k))) > eps) stop 1
```
mathrack, line 89: This looks like a bad idea !

mathrack (changes requested):

examples/fft_physical_x/fft_grid_x.f90
rfj82982: Since loop dimensions are all stored in arrays we have two options to run them on GPU: 1 convert loop start/end into scalars (the one used above) or move the array to device (used for example here) https://github.com/xcompact3d/2decomp-fft/blob/f9d047fcc998f3737cb3ce61c3e04848b12751ad/examples/test2d/test2d.f90#L48
I wanted to give example with both options but i understand that this look inconsistent. Open to suggestion to what is the best way.

examples/io_test/io_plane_test.f90
rfj82982: I add this note in the file to explain why i put an `#ifdef ` around some IO options. We should fix the IO routine but I before we need to agree what we want to do with IO and GPU.

examples/fft_physical_x/fft_grid_x.f90
mathrack: If the two solutions are equivalent, I think we have two options :

1. Use only the recommended one in the examples
2. Use both in the examples and explain somewhere there are two options

examples/io_test/io_test.f90
rfj82982: My mistake don’t know i substitute 2 with 1.

examples/io_test/io_plane_test.f90
mathrack: In my opinion, we should not modify the example. I suggest we have a dedicated issue and the following  : 

- Keep the current `call decomp_2d_write_plane` in the example
- Update the IO subroutines when writing plane

In case of GPU build, when writing planes, if transpose operations are needed and not supported, one can : 

- Print some error message and abort the simulation
- Print some warning message, skip the IO and run the simulation
- Silently skip the IO and run the simulation

rfj82982: What i have is already skipping the IO that it is not supported. I’ll add a warning. To fix the IO i would like to first check with Paul.

mathrack: My apologies if my previous comment was not clear. In my opinion, the tests performed in `io_plane_test.f90` should not be skipped for a GPU build.

The file `io.f90` or the file `io_write_plane.inc` should be updated instead. If we decide to print some error and abort, the modification should be

```
#if defined(_GPU)
   if (ipencil /= iplane) call decomp_2d_abort(__FILE__, &
                                                                               __LINE__, &
                                                                               ipencil, &
                                                                             "GPU build does not allow transpose operations when writing planes")
#endif
```

If we decide to print some warning and skip IO, the modification should be

```
#if defined(_GPU)
   if (ipencil /= iplane) then
      call decomp_2d_warning(__FILE__, &
                                                    __LINE__, &
                                                    ipencil, &
                                                    "GPU build does not allow transpose operations when writing planes")
      return
   endif
#endif
```

rfj82982: Yes we should update the IO files, however i think we need to check with Paul before going on modifying the files. I know how to fix the problems on the IO file, just need to check how this is fitting in the overall development for the IO.

rfj82982: Ok i think we can go for the second option of a warning and a return. I think the test needs to be completed since this is not really a bug but just a feature not being up to date. 
I’ll use the second option.

examples/fft_physical_x/fft_grid_x.f90
rfj82982: I think it is confusing to have both options. In my opinion we should not pass to GPU very small arrays like `xstat` or `xend`. I’ll modify all the examples having only local scalars as loop boundaries

examples/io_test/io_plane_test.f90
mathrack: I have pushed the warning + skip IO update. Only thing needed to resolve the conversation is to open a dedicated issue and clean the comments in examples/io_test/io_plane_test.f90

mathrack: Please keep in mind that the issue of plane IO for GPU builds can have a dedicated PR, it is not mandatory to solve it in this PR

examples/Makefile
```diff
@@ -1,36 +1,32 @@
...
 
+init_test:
+	$(MAKE) -C $@ all
 test2d:
 	$(MAKE) -C $@ $@
 fft_physical_x:
 	$(MAKE) -C $@ all
 fft_physical_z:
 	$(MAKE) -C $@ all
-halo_test:
```
mathrack, line 13: The test `halo_test` is removed. Why ?

examples/fft_physical_x/fft_grid_x.f90
rfj82982: I’m doing the examples with local variables. On Thursday we can decide a unique way forward and we can do all relative modifications.

examples/Makefile
rfj82982: I am finishing halo. This is why i have removed it. As soon as the code is working properly i’ll add it back

pbartholomew08: I think approach to test on atleast CPU is best. We can add GPU test later

examples/halo_test/halo_test.f90
```diff
@@ -334,6 +340,9 @@ subroutine test_div_haloZ()
 #else
       logical, parameter :: global = .false.
 #endif
+      integer :: i1, in ! I loop start/end
```
pbartholomew08, line 343: Note to self - check if `in` causes issues as a reserved keyword

src/fft_cufft.f90
```diff
@@ -408,24 +409,18 @@ subroutine c2c_1m_x(inout, isign, plan1)
...
 
-      complex(mytype), dimension(:, :, :), allocatable :: output
       integer :: istat
 
-      allocate (output, mold=inout)
 #ifdef DOUBLE_PREC
-      !$acc host_data use_device(inout,output)
-      istat = cufftExecZ2Z(plan1, inout, output, isign)
+      !$acc host_data use_device(inout)
+      istat = cufftExecZ2Z(plan1, inout, inout, isign)
```
pbartholomew08, line 416: Likewise - is `inout` a problem/reserved keyword?

pbartholomew08 (approved):

examples/halo_test/halo_test.f90
rfj82982: I also don’t like in. This was there from before and so far as not created any issue. I would be happy to change it, but we might have to change also also the rest accordingly for consistency i.e. i1, in j1, jn k1 kn. Shall we use double letter like ii1 iin?

src/fft_cufft.f90
rfj82982: inout is a bit everywhere and comes from first implementation? For the moment I’ll leave it there and have a PR where we change it consistently

examples/halo_test/halo_test.f90
pbartholomew08: `ifirst`/`ilast`? Pretty sure this is my fault so will make the change.

src/fft_cufft.f90
pbartholomew08: ok - sounds good, and compiler isn't complaining

PR #145, opened 2023-01-31
First porting to GPU of the halo code/example

rfj82982:


pbartholomew08:
Passing on Intel compilers

examples/halo_test/halo_test.f90
```diff
@@ -349,93 +393,73 @@ subroutine test_div_haloZ()
...
-      !      array seems to be OK.
       allocate (tmp(size(divh, 1), size(divh, 2), size(divh, 3)))
+
+      !$acc kernels default(present)
       tmp(2:xlast, 2:ylast, 2:zlast) = divh(2:xlast, 2:ylast, 2:zlast) - div1(2:xlast, 2:ylast, 2:zlast)
-      err = mag(tmp(2:xlast, 2:ylast, 2:zlast))
-      deallocate (tmp)
-      divmag = mag(div1(2:xlast, 2:ylast, 2:zlast))
+      !$acc end kernels
+      error = mag(tmp)
```
pbartholomew08, line 457: Passing wrong range?

examples/halo_test/halo_test.f90
```diff
@@ -349,93 +393,73 @@ subroutine test_div_haloZ()
...
-
       real(mytype), dimension(:, :, :), allocatable :: tmp
+      real(mytype) :: divmag, error
+#if defined(_GPU)
+      attributes(device) :: tmp
+#endif
 
-      real(mytype) :: divmag
-
-      ! XXX: The Intel compiler SEGFAULTs if the array difference is computed inplace
```
pbartholomew08, line 430: I think useful to keep comment in case someone tries to remove tmp array as it's not (obviously) required...

examples/halo_test/halo_test.f90
rfj82982: No the range is correct. In the original code they were allocating from 1, but when computing mag they were starting from 2 because of the -1 loop. I put the start to 2 in the loop to compute mag since the sum(tmp(:,:,:)) do not work on GPU since there is a reduction and this needs to be done with a loop.

examples/halo_test/halo_test.f90
```diff
@@ -452,16 +476,34 @@ subroutine check_err(divh, pencil)
...
+#if defined(_GPU)
+      attributes(device) :: a
+#endif
 
       real(mytype) :: lmag, gmag
 
-      lmag = sum(a(:, :, :)**2)
+      lmag = 0._mytype
+      !$acc parallel loop default(present) collapse(3) reduction(+:lmag)
+      do k = 2, zlast
```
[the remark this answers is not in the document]
rfj82982: Range should be the same just check my comment below.

src/halo_common.f90
```diff
@@ -114,47 +114,64 @@
...
+       !$acc kernels default(present)
        do k = 1, s3
           do j = 1, s2
              do i = 1, s1
                 out(i, j, k) = in(i, j, k)
              end do
           end do
        end do
+       !$acc end kernels
+       !istat = cudaMemcpy(out,in,s1*s2*s3,cudaMemcpyDeviceToDevice)
```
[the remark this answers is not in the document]
rfj82982: Yes this is just a pro-memoria since the all the loops above can be substituted with a cudaMemcpy. I wanted to keep the loop just to try and if we move to openMP with another compiler we have an option instead of a native CUDA function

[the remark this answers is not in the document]
pbartholomew08: Resolved by 2d8fb3f

examples/halo_test/halo_test.f90
pbartholomew08: Resolved by 5ab1f79

pbartholomew08 (approved):

PR #155, opened 2023-02-16
Remove the makefile

pbartholomew08:
Todo:
- [x] port CI to Cmake
- [x] add instructions for driving build/linking from an external Makefile


.github/workflows/Build.yml
```diff
@@ -24,11 +24,15 @@ jobs:
...
         sudo apt-get update
-        sudo apt-get install -y gfortran openmpi-bin libopenmpi-dev
+        sudo apt-get install -y gfortran openmpi-bin libopenmpi-dev cmake
 
     # Runs a set of commands using the runners shell
     - name: Compile
-      run: make BUILD=dev
+      run: |
+        FC=mpif90 cmake -B build -DCMAKE_BUILD_TYPE=dev
+        make -C build decomp2d
```
mathrack, line 33: Should be cmake ?

.github/workflows/Build_all.yml
```diff
@@ -99,15 +103,21 @@ jobs:
...
+      run: |
+        source /opt/intel/oneapi/setvars.sh
+        FC=mpiifort cmake -B build -DCMAKE_BUILD_TYPE=debug
+        make -C build decomp2d
 
     # Runs a set of commands using the runners shell
     - name: Compile
-      run: source /opt/intel/oneapi/setvars.sh && make clean && make CMP=intel FFT=mkl BUILD=debug && make examples
+      run: |
+        source /opt/intel/oneapi/setvars.sh
```
mathrack, line 118: Could be best to keep the make clean here ?

p-costa:
I guess that I will have to adopt `cmake` for building 2decomp if I want to keep my Git Submodule in sync with the upstream developments?

mathrack:
Hello @p-costa , this is an open-source project. We decided to switch to CMake some time age. Keeping up-to-date both Makefile and CMake leads to extra work...

However, a dedicated branch with support of Makefile could be maintained. If you want to take a share of this extra work, I am willing to contribute. Could be easier to maintain if restricted to CPU build, we can talk about it.

Cheers

p-costa:
Hi Cédric, @mathrack. Thanks!

I didn't want my query to come across as imposing or anything. Worst case, if I don't adopt `cmake`, I'll keep and maintain my own `make` workflow in my fork. Anyway, I'd be happy to discuss if you plan to keep `make` and `cmake` build workflow in parallel.

mathrack:
Sounds good, this is not merged yet, we will chat about it when it gets merged.

.github/workflows/Build.yml
pbartholomew08: Yes, we should follow the style from the README. Will fix

p-costa:
Is just that it will require some exercise on our end to integrate it in codes using CaNS as base. There the 2decomp build is connected to the build profile of CaNS.

.github/workflows/Build_all.yml
pbartholomew08: Addressed by fbc5660

.github/workflows/Build.yml
pbartholomew08: Addressed by 75bd117

pbartholomew08:
All tasks in CI have been updated to use cmake, I can run each task by hand (GNU build, Nvidia build, Intel build, MKL, FFTW3, FFTW3_F03 and caliper) and the tests pass so I assume they will continue to pass when the CI switches over with this merge (though that will be the true test)

mathrack (approved):

PR #156, opened 2023-02-17
12 update examples allowing user input for ctest

rfj82982:
Update of CTest to allow the user to specify decomposition and resolution at runtime and configute stages.


rfj82982:
> Some code is duplicated, this is not very convenient.

I’ll try to see if i can find a more elegant solution to pass the inputs.

mathrack:
@rfj82982 you can merge and look for a more elegant solution later if you want

PR #159, opened 2023-03-02
Fix nccl type

rfj82982:
Fix the issue with nccl Type. Tested in both single and double precision on 4 GPUs


pbartholomew08 (approved):
LGTM!

PR #160, opened 2023-03-14
Module cumpi

rfj82982:
First update to separate the CUDA aware MPI part of the decomp2d module. 
Part of MPI is also included into constants. This should also become a stand alone module.


src/decomp_2d_init_fin.f90
```diff
@@ -47,34 +47,12 @@ subroutine decomp_2d_init_ref(nx, ny, nz, p_row, p_col, periodic_bc, comm)
...
 
-     ! Use the provided MPI communicator if present
-     if (present(comm)) then
-        decomp_2d_comm = comm
-     else
-        decomp_2d_comm = MPI_COMM_WORLD
-     end if
+     call decomp_2d_mpi_init(comm)
 
-     ! Safety check
```
mathrack, line 57: This was removed, it should be harmless to keep it

mathrack (changes requested):
The code looks ready for merging. I only suggest to restore the comparison between MPI_SUCCESS and 0 inside `decomp_2d_mpi_init`

src/decomp_2d_init_fin.f90
rfj82982: Yes I’ll put it back, but in `decomp_2d_mpi_init` under the `constants` module.

src/decomp_2d_constants.f90
```diff
@@ -81,5 +121,251 @@ module decomp_2d_constants
...
+#if defined(_GPU) && defined(_NCCL)
+      module procedure decomp_2d_abort_nccl_basic
+      module procedure decomp_2d_abort_nccl_file_line
+#endif
+   end interface decomp_2d_abort
+
+   interface decomp_2d_warning
+      module procedure decomp_2d_warning_basic
+      module procedure decomp_2d_warning_file_line
+   end interface decomp_2d_warning
```
pbartholomew08, line 142: The constants module seems like a strange place for these?

pbartholomew08:
There seems to be a lot of functionality inside the constants module, this doesn't make sense to me - it should contain constants like pi, maybe "constants" like the grid size I think but not methods for initialising MPI, raising warnings, etc.

mathrack:
It makes sense. All the variables in decomp_2d_constants could have the parameter attribute. And we could also switch the 2 enum to `integer, parameter`. However, this could be done in a dedicated PR.

Maybe this PR could be limited to cuda-aware MPI stuff and leave untouched the MPI parts.

mathrack:
I suggest the following : https://github.com/rfj82982/2decomp-fft/pull/4

mathrack:
The only issue is that the module decomp_2d_mpi contains the stuff for errors and warnings. But the stuff for errors and warnings needs nrank from the mpi module

rfj82982:
I put everything under constants since there were some circular dependancies. i had a `decomp_2d_mpi` module, but the file was constantly compiled, under CMake, after the main decomp_2d module for which it is a dependancy. My idea was to do all the modifications and after look at why i had that problem and properly have an MPI module. I’ll have a look at Cedric suggestion and check it the issue with CMake is still there.

rfj82982:
I think this is ready to be merged. We should check also with Intel since with gnu everything was OK, only NVHPC highlighted some missing modules.

mathrack (approved):

pbartholomew08 (approved):

PR #161, opened 2023-03-14
Module nccl

rfj82982:
Separation of the NCCL code from decomp2d module. This has a first implementation of the transpose for complex using NCCL. This PR should be included after the cuMPI (#160). Further work is on going to improve the transpose operation for complex.

PR #162, opened 2023-03-14
Update transpose for slabs

rfj82982:
Third PR to perform a memory copy rather than a transpose operation via communication (MPI/NCCL) in case of slabs.


pbartholomew08:
Marking as draft as I think dependent on #160 (and maybe #161?)

rfj82982:
This should be the last PR to have decomp2d ready to be released.

mathrack (approved):

mathrack:
I will merge. A last PR is needed to prettify the code before release.

rfj82982:
OK no problem. To be complete we should look into pointers to do complex transpose, but we might do this more generally and save about 1/3 of the memory footprint.

mathrack:
@rfj82982 there is also some potential for memory saving using in the FFT engine :

- [in-place r2c and c2r](https://github.com/xcompact3d/2decomp-fft/pull/144) transforms
- reduced memory requirement if only r2c / c2r transforms are used and no c2c transform is needed

Issue #168, opened 2023-03-20
Include major and minor version number in the log


mathrack:
Currently, the library can print in the log the hash of the last git commit : https://github.com/xcompact3d/2decomp-fft/blob/main/src/log.f90#L143

This is relevant for a development version of the library. When a stable version is released, the git hash should be replaced or completed by the major / minor version.

PR #169, opened 2023-03-20
Print major and minor version number

mathrack:
Closing #168


pbartholomew08 (approved):

pbartholomew08:
Do we remove the Git Version macro?

mathrack:
In my opinion, it is good to keep the Git version macro for development builds. But if the preprocessor variable VERSION is not defined, it is not needed to print "unknown" ?

https://github.com/xcompact3d/2decomp-fft/blob/version_number/src/log.f90#L150

pbartholomew08:
ok, seems it's not causing anyone else issues

PR #172, opened 2023-03-31
170 add cmake install packaging

pbartholomew08:
This will allow other packages using 2decomp to use the `find_package(decomp2d)` functionality.


rfj82982:
I have just modified the README. I have tested the linking as an external library with this version of X3div 
https://github.com/rfj82982/x3div/tree/2d-decomp-fft_external_library
I’ll try to do more testing also with GPU before approving.

rfj82982 (approved):
LGTM

PR #186, opened 2023-04-26
Release v2

pbartholomew08:


rfj82982 (dismissed):
I have review all the changes and tested also on GPUs. Only issue is the latest NCCL version (V2.16) with CUDA12. Version with CUDA MPI is OK. I’ll look at this issue after the merging.

pbartholomew08:
I think it is probably good to go

rfj82982:
Ok I’ll have a look in about 10 min

rfj82982 (approved):

PR #191, opened 2023-05-02
Update CMake to enable the EVEN preprocessor flag

mathrack:
Update the code to hide EVEN-specific variables

### DIFF
--- a/src/decomp_2d.f90
+++ b/src/decomp_2d.f90
@@ -100,12 +100,12 @@ module decomp_2d
       integer, allocatable, dimension(:) :: &
          x1disp, y1disp, y2disp, z2disp
 
-      ! buffer counts for MPI_ALLTOALL: either for evenly distributed data
-      ! or for padded-alltoall
+#ifdef EVEN
+      ! buffer counts for MPI_ALLTOALL for padded-alltoall
       integer :: x1count, y1count, y2count, z2count
-
       ! evenly distributed data
       logical :: even
+#endif
 
    END TYPE DECOMP_INFO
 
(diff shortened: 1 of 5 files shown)


mathrack:
A side question connected with this PR : do we want the CI to test compilation for the cases below ?

- EVEN preprocessor variable ? (EVEN CMake option)
- OVERWRITE preprocessor variable ? (ENABLE_INPLACE CMake option)
- HALO_DEBUG preprocessor variable ? (HALO_DEBUG CMake option)

A complete CI should test the OVERWRITE preprocessor variable for the various FFT backends.

src/decomp_2d.f90
```diff
@@ -406,10 +399,18 @@ subroutine decomp_info_init(nx, ny, nz, decomp)
...
       buf_size = max(buf_size, &
                      max(decomp%x1count * dims(1), decomp%y2count * dims(2)))
+      ! evenly distributed data ?
+      if (mod(nx, dims(1)) == 0 .and. mod(ny, dims(1)) == 0 .and. &
+          mod(ny, dims(2)) == 0 .and. mod(nz, dims(2)) == 0) then
+         decomp%even = .true.
+      else
+         decomp%even = .false.
+      end if
 #endif
```
pbartholomew08, line 414: I would add a 
```
#else 
  decomp%even = .false.
#endif
```
to ensure there is a default value in case it is tested elsewhere.

pbartholomew08 (changes requested):
Looks good, only a minor suggestion

src/decomp_2d.f90
mathrack: the variable does not exist if the preprocessor variable is not present

pbartholomew08: ah, ok

pbartholomew08 (approved):

mathrack:
@pbartholomew08 what do you think about the comment above on CI coverage ?

PR #196, opened 2023-05-10
Fix examples for single precision builds with dev flags

mathrack:


examples/fft_physical_x/fft_c2c_x.f90
```diff
@@ -160,14 +160,14 @@ program fft_c2c_x
       write (*, *) '===== c2c interface ====='
       write (*, *) 'error / mesh point: ', err_all
       write (*, *) 'time (sec): ', t1, t3
-      n1 = real(nx, mytype) * real(ny, mytype) * real(nz, mytype)
-      n1 = n1**(1._mytype / 3._mytype)
+      n1 = real(nx) * real(ny) * real(nz)
```
pbartholomew08, line 163: Should be `dbl(nx)`, etc. I think

pbartholomew08 (changes requested):

examples/fft_physical_x/fft_c2c_x.f90
mathrack: `real` will promote to the default floating point type, this could be float or double. Worst case scenario, a few digits are lost when estimating the FLOPS. I can fix it tomorrow if this is a major issue.

pbartholomew08: I don't think the digits matter, given the issue with type conversions I thought this would give a similar issue (`n1` is `double precision`) but testing locally (with OpenMPI + gfortran 9/10) this appears not to be a problem.

pbartholomew08 (approved):

PR #197, opened 2023-05-11
Update Build_all.yml

mathrack:
Update the CI script to restore the default flags

### DIFF
--- a/.github/workflows/Build_all.yml
+++ b/.github/workflows/Build_all.yml
@@ -37,32 +37,32 @@ jobs:
     - name: Single precision
       run: |
         export FC=mpif90.openmpi
+        rm -rf ./build
         cmake -S . -B build -DCMAKE_BUILD_TYPE=Dev -DBUILD_TESTING=ON -DDOUBLE_PRECISION=OFF
-        cmake --build build --target clean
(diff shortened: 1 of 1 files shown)


.github/workflows/Build_all.yml
```diff
@@ -145,17 +145,17 @@ jobs:
         source /opt/intel/oneapi/setvars.sh
         export FC=mpiifort
```
pbartholomew08, line 146: also `rm -rf ./build`?

.github/workflows/Build_all.yml
```diff
@@ -178,11 +178,11 @@ jobs:
         export FC=mpif90
```
pbartholomew08, line 178: also `rm -rf ./build`?

mathrack: For the first build of the step, no cleaning is needed

.github/workflows/Build_all.yml
mathrack: For the first build of the step, no cleaning is needed

pbartholomew08:
ok, wasn't sure if they necessarily ran in order

pbartholomew08 (approved):

PR #207, opened 2023-05-27
Increase the size of the name of the log file

mathrack:
The file returned can contain the absolute path and can be relatively long.

### DIFF
--- a/src/log.f90
+++ b/src/log.f90
@@ -116,7 +116,7 @@ contains
       integer :: io_unit
       integer :: version, subversion, ierror
 #ifdef DEBUG
-      character(len=64) :: fname
+      character(len=512) :: fname
 #endif
 
       ! Output log if needed


src/log.f90
```diff
@@ -116,7 +116,7 @@ module subroutine d2d_listing(given_io_unit)
       integer :: io_unit
       integer :: version, subversion, ierror
 #ifdef DEBUG
-      character(len=64) :: fname
+      character(len=512) :: fname
```
pbartholomew08, line 119: We could use an allocatable string to resolve this issue.
```
character(len=:), allocatable :: fname

fname = "/really/really/long/path/to/log.log" ! Auto-allocated to fit
```

pbartholomew08 (changes requested):

mathrack:
I am not sure it is possible because the name is obtained with

```
inquire (unit=io_unit, name=fname, iostat=ierror)
```

May 27, 2023 19:30:48 Paul Bartholomew ***@***.***>:

pbartholomew08:
Ah, ok

pbartholomew08 (approved):

src/log.f90
pbartholomew08: Name obtained via `inquire`, needs a fixed size

PR #209, opened 2023-06-13
Make determining MPI compilers more robust

pbartholomew08:
As a fallback option, use the -show argument to MPI to try and find the MPI family.

Default is "Unknown"

Currently this is only important for gfortran > 10 and dev builds with only OpenMPI supplying the necessary modules to turn up the warning/error flags fully. If the MPI family is MPICH or Unknown then less strict flags are used.


rfj82982 (dismissed):
I tested and it seems ok

rfj82982 (approved):

Issue #211, opened 2023-06-23
Intel Compiler


rfj82982:
**Describe the bug**
There are sone minor issues with the Intel compiler mainly related to the new IntelLLVM. To compile and run flags needs to be differentiate between the two compilers. 
The ifx has also the following bug: 
- sqrt(1.0) return 0.9999 instead of 1 and this create an issue in serial since the decomposition routine do not  return a 1x1 decomposition but a rather 0xLargeInt

Additionally some safety checks design to check the existence of directories before running a test using the Fortran function `inquire` have a different behaviour than with other compilers returning always a `false` flag even if the directory exists. Intel fortran `inquire` functionality return `true` only if a file exists.

pbartholomew08:
Can this be closed now?

rfj82982:
yes

PR #212, opened 2023-06-23
Fix intel

rfj82982:
This PR is to address the issues with the Intel compiler raised in issue #211. 
The PR provides: 
- different flags for ifort and ifx compilers
- a fix for the IntelLLVM fortran compiler to have the correct result for the sqrt(1.0) by forcing the integer conversion to double and not single precision
- a different `inquire ` argument pointing to a file instead of a directory. 

Test have been performed using:
- IntelLLVM 2022.2.1
- ifort 2021.7.0.20221019
The ifort version testing has been done also using ADIOS2 for the IO backhand both in serial and parallel (up to 4 mpi rank). Both version of the Intel Fortran have been tested with the generic FFT and with the MKL.

### DIFF
--- a/examples/io_test/io_read.f90
+++ b/examples/io_test/io_read.f90
@@ -34,7 +34,7 @@ program io_read
 
    character(len=*), parameter :: io_name = "test-io"
 #ifndef ADIOS2
-   logical ::file_exists
+   logical ::file_exists1, file_exists2, file_exists3
 #endif
 
    integer :: i, j, k, m, ierror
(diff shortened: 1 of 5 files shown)


examples/io_test/io_read.f90
```diff
@@ -91,8 +91,8 @@ program io_read
 
 #ifndef ADIOS2
    if (nrank == 0) then
-      inquire (file="out", exist=dir_exists)
-      if (.not. dir_exists) then
+      inquire (file="out/u1.dat", exist=file_exists)
```
pbartholomew08, line 94: Ah, I hadn't realised this was the test the issue was in. I think we should check existence of all the files to be read in.

pbartholomew08 (changes requested):
I'll add the code for my comment.

rfj82982 (commented):
Ok

pbartholomew08 (approved):

PR #217, opened 2023-06-27
Add adios2 git lab actions

rfj82982:


mathrack:
I think this will not work due to limitations on memory or disk space.

I would recommend to pull a docker image instead. See https://github.com/tlestang/test_adios_docker/blob/main/.github/workflows/test.yml

mathrack (changes requested):
See comment on memory and disk space

rfj82982:
This PR is closed since the testing of ADIOS within the CI will be done on the Actions/Runner and not with the Actions on GitHub

PR #219, opened 2023-07-02
Minor refactoring in transpose subroutines

mathrack:
When buffers are passed as arguments of the transpose subroutine, some duplicated code can be removed.

The modification is currently only applied to the transpose_x_to_y but should be applied to the other files if the PR is relevant


mathrack:
Closing. Will re-open for merging against branch dev.

PR #226, opened 2023-07-22
Test restarts

pbartholomew08:
Restart-type I/O (multiple fields per file) wasn't being tested. Extended `io_test` to do file per field and single file, multiple field (i.e. checkpoint) writes and reads. This should be added to any branch used for JOSS


rfj82982 (approved):
LGTM

PR #227, opened 2023-07-24
Test restarts

rfj82982:


pbartholomew08 (approved):
This is just adding the fix CI typo right?

rfj82982:
This is your update on the example/io where i added the typo correction. I had to merge the dev since there was the conflict with the major/minor parameter definition

src/io.f90
```diff
@@ -449,21 +449,21 @@ subroutine adios2_read_one_real(var, engine_name, varname, io_name)
 
       call adios2_variable_steps(nsteps, var_handle, ierror)
       if (ierror /= 0) call decomp_2d_abort(__FILE__, __LINE__, ierror, "adios2_variable_steps")
-      print *, "AVAILABLE steps for ", nsteps
+      ! print *, "AVAILABLE steps for ", nsteps
```
mathrack, line 452: Why do we keep this code commented ? It is probably best to use it only for debug builds ?

PR #244, opened 2023-08-09
Timing examples

rfj82982:
Update the timing of the example by giving results over an average and excluding the first iteration.

### DIFF
--- a/examples/fft_physical_z/fft_r2c_z.f90
+++ b/examples/fft_physical_z/fft_r2c_z.f90
@@ -110,12 +110,27 @@ program fft_r2c_z
       end do
    end do
 
-   t2 = 0.d0
-   t4 = 0.d0
    !$acc data copyin(in_r) copy(out)
    ! First iterations out of the counting loop
+   t1 = MPI_WTIME()
    call decomp_2d_fft_3d(in_r, out)
+   t2 = MPI_WTIME() - t1
+   t3 = MPI_WTIME()
    call decomp_2d_fft_3d(out, in_r)
+   t4 = MPI_WTIME() - t3
+   call MPI_ALLREDUCE(t2, t1, 1, MPI_DOUBLE_PRECISION, MPI_SUM, &
+                      MPI_COMM_WORLD, ierror)
+   t1 = t1 / dble(nproc) 
+   call MPI_ALLREDUCE(t4, t3, 1, MPI_DOUBLE_PRECISION, MPI_SUM, &
+                      MPI_COMM_WORLD, ierror)
+   t3 = t3 / dble(nproc) 
+   if (nrank == 0) then
+      write (*, *) '===== r2c/c2r interface ====='
+      write (*, *) 'It 0 time (sec): ', t1, t3
+   endif
+   ! Init the time
+   t2 = 0.d0
+   t4 = 0.d0
    !$acc kernels
    in_r = in_r / (real(nx, mytype) * real(ny, mytype) * real(nz, mytype))
    !$acc end kernels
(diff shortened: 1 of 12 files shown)


examples/halo_test/halo_test.f90
```diff
@@ -618,7 +618,7 @@ subroutine test_halo_size(arrh, nx_expected, ny_expected, nz_expected, tag)
 
          all_pass = .false.
       else
-         write (*, *) trim(rank_lbl), " ", tag, ":PASS"
+         !write (*, *) trim(rank_lbl), " ", tag, ":PASS"
```
mathrack, line 621: Keep the line or remove it. If needed, keep it and skip it when the debug mode is not active.

mathrack (changes requested):
I thing it is not a very good idea to skip the first iteration and then average the remaining ones. I would recommend to time all the iterations. And print all the timers at the end. To get a synthetic number, one could then drop the slowest and fastest timings, or the outliers.

examples/halo_test/halo_test.f90
rfj82982: I still think it is best to remove the first iteration since it tend to skew the results. After it iteration time tends to be relatively constant and the 50 iterations smooth the ups and down. 
In practice you’ll not see the effect of the first iteration in a simulation, but to take it out we’ll need around 100 iterations which is uncessary and because the large number of tests we are doing  start to make the testing a bit too long.

mathrack (changes requested):
I recommend to time all the iterations. At the end of the test, I think debug builds could print all the timers. Regardless of the build type, at the end of the test, the code could print min / max / average / average without the first iteration.

pbartholomew08:
Won't the first iteration always be slowest due to the FFT setup time? If we wanted to get some statistics we could do that across all iterations for iteration 2+, but including the first in those statistics I think will tend to skew them. We could also time the first iteration independently to get some measure of the cost of the setup.

mathrack:
I think most of the setup is done in decomp_2d_fft_init

examples/fft_physical_z/fft_r2c_z.f90
```diff
@@ -122,8 +131,9 @@ program fft_r2c_z
       call decomp_2d_fft_3d(out, in_r)
       t4 = t4 + MPI_WTIME() - t3
 
+      ! normalisation - note 2DECOMP&FFT doesn't normalise
       !$acc kernels
-      in_r = in_r / real(nx, mytype) / real(ny, mytype) / real(nz, mytype)
+      in_r = in_r / (real(nx, mytype) * real(ny, mytype) * real(nz, mytype))
```
pbartholomew08, line 136: From experience it's better to do the triple division as the product can overflow for large grids (although this may not happen with the `real` conversion)

pbartholomew08 (commented):
LGTM!

src/factor.f90
```diff
@@ -23,8 +23,8 @@ subroutine findfactor(num, factors, nfact)
 
       ! find the factors <= sqrt(num)
       ! Cast the int as double to make sure of the correct result of sqrt
-      ! IntelLLVM got an issue with 1.0 but not with 1.d0 
-      m = int(sqrt(real(num,8)))
+      ! IntelLLVM got an issue with 1.0 but not with 1.d0
```
pbartholomew08, line 26: I think the comment is now out of date? And `real(num, 8)` isn't portable, suggest `real(num, kind(1.0d0))` or `real(num, mytype)`

pbartholomew08 (changes requested):

src/factor.f90
mathrack: Should be fixed in https://github.com/2decomp-fft/2decomp-fft/pull/244/commits/1b085bd668d5c8f49a7f433dfa4ae15f961d3ddd.

pbartholomew08 (approved):

PR #246, opened 2023-08-15
245 fortran submodule use not working with aocc compiler

pbartholomew08:


pbartholomew08:
This replaces using `ph` and `sp` directly in the `fft_log` with calls to `get_ph/sp` which appears to resolve the AOCC compilation issue. Need to check it runs, and also compiles under other compilers.

pbartholomew08:
It appears to run ok at least in serial and seems not to have broken CI

mathrack (approved):

ningli-amd:
I notice that the v2.0.1 release last month does not contain fixes in this 4-moth-old PR therefore I still cannot build the library using AOCC. I wonder why.

rfj82982:
I'll have a look but potentially this will go into dev not the main

pbartholomew08:
It's already merged into and available on `dev` - it happened after we started preparing the 2.0.1 release

ningli-amd:
Thank you for you quick response. I am looking into providing Spack support for the library and the patch in #307 can be easily applied to the recipe for v2.0 and v2.0.1 releases. When is your next release scheduled?

mathrack:
I think we will release something soon because the nccl transpose subroutines were fixed not long ago (https://github.com/2decomp-fft/2decomp-fft/pull/306). Not yet scheduled though. We could release after merging #307 ?

rfj82982:
Yes we can do this

rfj82982:
i'll have a look at releasing v2.0.2

PR #251, opened 2023-09-01
Profile transpose operations when no MPI communication is performed

mathrack:
Current transpose operations are not always profiled when p_row or p_col is equal to one.

### DIFF
--- a/src/transpose_x_to_y.f90
+++ b/src/transpose_x_to_y.f90
@@ -24,6 +24,10 @@
      integer :: istat, nsize
 #endif
 
+#ifdef PROFILER
+     if (decomp_profiler_transpose) call decomp_profiler_start("transp_x_y_r")
+#endif
+
      if (dims(1) == 1) then
 #if defined(_GPU)
         nsize = product(decomp%xsz)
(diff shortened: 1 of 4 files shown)


src/transpose_x_to_y.f90
```diff
@@ -51,10 +59,6 @@ subroutine transpose_x_to_y_real(src, dst, decomp)
      integer :: s1, s2, s3, d1, d2, d3
      integer :: ierror
 
-#ifdef PROFILER
-     if (decomp_profiler_transpose) call decomp_profiler_start("transp_x_y_r")
-#endif
-
```
pbartholomew08, line 57: Will this not now miss calls to transpose routines when using the full interface, e.g. in the divergence/Poisson routines in Xcompact3d?

Would it be better to restore the original profiler and add a special case "transp_x_y_memcpy" for the special case when a copy is used instead of communication?

mathrack: I think this is ok. The full interface is not available outside of the library.

mathrack: See https://github.com/2decomp-fft/2decomp-fft/blob/always_profile_transpose/src/decomp_2d.f90#L188

pbartholomew08: It's not? Then these will break https://github.com/xcompact3d/Incompact3d/blob/34d69d2170bc172faafbdf80f697298ad9e89d0a/src/navier.f90#L317

pbartholomew08: oh, I see, my confusion.

pbartholomew08 (approved):

PR #263, opened 2023-09-08
Visu example

pbartholomew08:
Adds a visualisation example, still need to check visualising a complex field (xdmf doesn't support complex fields).

Also should we pull into SubmissionJOSS?

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -12,3 +12,6 @@ add_subdirectory(fft_physical_x)
 add_subdirectory(fft_physical_z)
 add_subdirectory(halo_test)
 add_subdirectory(io_test)
+
+# Set real/complex tests
+set(COMPLEX_TESTS "OFF" CACHE STRING "Enables complex numbers for tests that support it")
--- a/examples/io_test/CMakeLists.txt
+++ b/examples/io_test/CMakeLists.txt
@@ -3,6 +3,7 @@ file(GLOB files_read io_read.f90)
 file(GLOB files_var_test io_var_test.f90)
 file(GLOB files_plane_test io_plane_test.f90)
 file(GLOB files_bench io_bench.f90)
+file(GLOB files_visu visu.f90)
 
 include_directories(${CMAKE_SOURCE_DIR}/src)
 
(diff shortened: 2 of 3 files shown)


mathrack:
Maybe a file `adios2_config.xml` is needed ?

examples/CMakeLists.txt
```diff
@@ -12,3 +12,6 @@ add_subdirectory(fft_physical_x)
 add_subdirectory(fft_physical_z)
 add_subdirectory(halo_test)
 add_subdirectory(io_test)
+
+# Set real/complex tests
+set(COMPLEX_TESTS "OFF" CACHE STRING "Enables complex numbers for tests that support it")
```
mathrack, line 17: Currently, the test is :

- Default : off, test with real only
- On, test with complex only

I think it is best to test both real and complex variables when the test is on

rfj82982: I am having a look at this

mathrack:
I think we could remove the complex part in this PR and merge it in both main and dev. We can add some tests for complex IO in dev later.

examples/CMakeLists.txt
rfj82982: I would like to merge this. What is the consensus? Do we do a test only for real, real/complex(current version) or real and complex?   There are quite a bit of `#ifdef` i would like to move the test to have it only for `#ifdef ADIOS2`

pbartholomew08:
I think the consensus is to just test `real` for now, will remove the complex bits

rfj82982:
HI Paul I can do it since I want to add a couple of other stuff to this PR. I'll let you know when done (I have to first check a very unhappy french)

pbartholomew08:
I've updated the `adios2_config.xml` for `io_visu`, it requires HDF5 for the xdmf reader.

mathrack:
This can be merged into main. To merge into dev, it would be best to use `decomp_2d_testing_init`.

rfj82982 (approved):
Now LGTM, we have to remember just to always use native BP4 with ADIOS2 and avoid HDF5 to remove the extra dependency

PR #264, opened 2023-09-11
Some code is duplicated in the examples. Move it to a dedicated module

mathrack:


pbartholomew08:
My suggestion would be to put this in something like `examples/utilities.f90` - otherwise we risk imposing an argument count/ordering on users (who may have other arguments they need to pass).

mathrack:
Currently, the processing of the command-line arguments ships inside the library 2decomp. It is present in the library only when it is compiled with the tests.

It makes sense to move the file outside of s`src` and to put it inside a folder such as `examples/utilities` or similar.

Are you also suggesting to move it outside of the library 2decomp ?

rfj82982:
I am in favour of something inside `examples`. I do not want an extra module in `src` since here we should have only 2DECOMP&FFT mandatory modules/code.

pbartholomew08:
I would make it not part of the library, I would have it as a utilities module that is compiled as part of the test suite only

mathrack:
This is ready for review

pbartholomew08:
I think we can use a cmake interface library (not sure if that's exactly the name) - I can take a look later (probably tomorrow) at doing this.

rfj82982:
I’ll have a look as well

pbartholomew08:
I have a fix that builds in parallel locally in #265

pbartholomew08:
> I think we can use a cmake interface library (not sure if that's exactly the name) - I can take a look later (probably tomorrow) at doing this.

The CMake terminology is "Object Library" - it allows you to bundle up object files and treat them logically like a library so that the correct dependencies can be specified.

pbartholomew08 (approved):
I think this looks good assuming the CI passes

pbartholomew08:
Looks like this might have been merged prematurely, failing on the MPICH build test
```
/usr/bin/mpif90.mpich -cpp -std=f2008 -ffree-line-length-none -fallow-argument-mismatch -DDEBUG -g3 -Og -ffpe-trap=invalid,zero -fcheck=all -fimplicit-none -Wall -Wno-unused-function -Wno-integer-division CMakeFiles/init_test.dir/init_test.f90.o ../CMakeFiles/examples_utils.dir/utilities.f90.o -o ../../bin/init_test  ../../lib/libdecomp2d.a ../../lib/libdecomp2d.a 
```
I reckon due to twice linking `libdecomp2d.a`

PR #265, opened 2023-09-12
Link the examples utilities as an object library

pbartholomew08:
CMake object libraries are "logical" libraries that allow us to define a set of object files that depend on another library, here

example_utils depends on decomp2d
example/X depends on decomp2d and example_utils

Parallel build now succeeds (locally) with `-j4`


mathrack (approved):

rfj82982:
Ok I was about to push something similar. I had just an extra directory and doing something similar to Paul

rfj82982:
Sorry about this but I had to attend the OPS meeting. Starting to split into 2

Issue #33, opened 2022-08-11
Re-design IO logic


mathrack:
Some progress on this issue is possible after issue [https://github.com/xcompact3d/2decomp-fft/issues/9](https://github.com/xcompact3d/2decomp-fft/issues/9) is solved.

The proposed IO logic would rely on a list of readers and writers.

### A reader / writer

1. Would perform IO using MPI, ADIOS2, raw binary or text (only one). MPI and ADIOS2 correspond to parallel IO. Raw binary and text correspond to serial IO. ADIOS2 could also correspond to serial IO.
2. Can do IO using single or double precision but not both (whatever the value of the variable `mytype`)
3. Can do IO for real and complex variables, integers and logicals
4. Can do parallel IO only for fields associated with a given decomp_info object
5. Can do parallel IO for 3D scalars and 2D slices
6. Can do serial IO for 1D profiles and 0D probes
7. Can perform transpose operations in case of parallel IO with native precision
8. Can perform MPI operations in case of serial IO

### A reader object should include the following information

1. Read at the beginning of the simulation ?
2. Read at a given frequency ? (probably at the beginning of the time step, but this is out of the scope of decomp_2d)
3. Id of the reader (unique)
4. Name of the reader (unique)
5. Directory for the reader
6. Optional, number of files / variables the reader is associated with (a reader can read several files / variables)
7. Optional, names of the files / variables the reader is associated with (a reader can read several files / variables)
8. If parallel IO, associated decomp_info object

### A writer object should include the following information

1. Write at the beginning of the simulation ?
2. Write at the end of the simulation ?
3. Write at a given frequency ? (probably at the end of the time step, but this is out of the scope of decomp_2d)
4. Id of the writer (unique)
5. Name of the writer (unique)
9. Directory for the writer
10. Optional, number of files / variables the writer is associated with (a writer can write several files / variables)
11. Optional, names of the files / variables the writer is associated with (a writer can write several files / variables)
12. If parallel IO, optional, is-it coarse-grained ?
13. If parallel IO, associated decomp_info object

pbartholomew08:
I think this will be good to discuss in person in September, before we do any changes like this I'd like to do some tidying (e.g. lift all the MPIIO stuff into an `io_mpiio` submodule and likewise ADIOS2 into `io_adios2` submodule) this should make it easier for the two to work together and clean up the code to make significant logic changes easier to implement.

mathrack:
[...]

PR #269, opened 2023-09-22
Refactor io

pbartholomew08:
Opening so we can discuss the changes - relates to #33


src/alloc.f90
```diff
@@ -4,22 +4,67 @@
   !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
 
   ! X-pencil real arrays
-  subroutine alloc_x_real_short(var, opt_global)
+  subroutine alloc_x_freal_short(var, opt_global)
 
      implicit none
 
-     real(mytype), allocatable, dimension(:, :, :) :: var
+     real(kind(0._real32)), allocatable, dimension(:, :, :) :: var
```
pbartholomew08, line 11: I think you should only need to write `real(real32)` and similar for the double case below

src/alloc.f90
```diff
@@ -217,7 +438,7 @@ subroutine alloc_z_real(var, decomp, opt_global)
      if (present(opt_global)) then
         global = opt_global
      else
-        global = .false.
+        global = default_opt_global
```
pbartholomew08, line 441: This is potentially a confusing or significant change from how the code works now - would it be more appropriate to do in v3? (v2.1 shouldn't change behaviour or ideally interface over v2).

src/decomp_2d_constants.f90
```diff
@@ -100,6 +100,27 @@ module decomp_2d_constants
...
+   integer, parameter, public :: DECOMP_2D_WRITE_MODE = 1, &
+                                 DECOMP_2D_READ_MODE = 2, &
+                                 DECOMP_2D_APPEND_MODE = 3
+
+   integer, parameter, public :: DECOMP_2D_WRITE_DEFERRED = 1, &
+                                 DECOMP_2D_WRITE_SYNC = 2
+   !
+   ! Interpolation methods
+   !
+   integer, parameter, public :: DECOMP_2D_INTERP_BASIC = 0
```
pbartholomew08, line 122: Is basic 2nd order or using the compact interpolation schemes? Could add a comment

src/io_family.f90
```diff
@@ -0,0 +1,586 @@
...
+   public :: decomp_2d_io_family_get_default_adios, &
+             decomp_2d_io_family_set_default_adios
+#endif
+
+contains
+
+   !
+   ! Initialize a new family of readers / writers (default type)
+   !
+   subroutine d2d_io_family_init(family, label)
```
pbartholomew08, line 61: Could this receive an (optional) argument to request a specific backend? e.g.
```
call d2d_io_family_init(family, label, backend=D2D_ADIOS2_BACKEND)
```

src/io_family.f90
```diff
@@ -0,0 +1,586 @@
...
+#endif
+
+      ! Safety check
+      if (family%type /= decomp_2d_io_none) then
+         call decomp_2d_abort(__FILE__, __LINE__, -1, "Family was not cleared "//label)
+      end if
+
+#ifndef ADIOS2
+      call decomp_2d_abort(__FILE__, __LINE__, -1, "ADIOS2 is not available")
+#else
```
pbartholomew08, line 131: I'd put this at the top of the subroutine, no point doing anything further if ADIOS2 isn't available when this is called.

src/alloc.f90
mathrack: The parameter `default_opt_global` is set to false in the PR (inside decomp_2d.f90). My intention is to have only a single parameter for all the alloc subroutines instead of hard-coded values duplicated

src/io_family.f90
mathrack: The current design allow the external code to select the backend using `family%mpi_init` or `family%adios_init`. The addition of one optional parameter for calls to `family%init` is a simple addition and I will implement it. The allowed values for the optional argument will be `DECOMP_2D_IO_MPI` and `DECOMP_2D_IO_ADIOS2`.

src/decomp_2d_constants.f90
mathrack: I think the method is order zero (closest neighbor). I will provide some comments.

Please note that the addition of the interpolation is connected with the removal of the parameters icoarse / nstat/  nvisu / nprobe.

src/alloc.f90
mathrack: This will make the code easier to read. I will push it later.

src/decomp_2d_constants.f90
mathrack: Fixed in https://github.com/2decomp-fft/2decomp-fft/pull/269/commits/4109f442f1a3c60bf7b87c9cc92ddca372e65981

src/io_family.f90
mathrack: Fixed in https://github.com/2decomp-fft/2decomp-fft/pull/269/commits/482a23bdc19b8877906f901558c2cd466c921b36

src/io_family.f90
mathrack: https://github.com/2decomp-fft/2decomp-fft/pull/269/commits/ce1aeabac86595c5204cf2a89b3884fecd086f26

src/alloc.f90
mathrack: As a side remark, the new design will change the IO interface.

mathrack:
CI is working with GFortran. I am not sure to understand the NVHPC error.

mathrack:
The CI is now working. The examples do not cover 100% of the capabilities for the new IO but this is a good starting point.

I have made some design choices when sketching this. This is not the final draft, feel free to discuss my sketch.

If needed, I can update the Changelog to describe the modification.

examples/io_test/io_bench.f90
```diff
@@ -55,6 +66,7 @@ program io_bench
    end if
 
    deallocate (u1)
+   call decomp_2d_io_fin
```
pbartholomew08, line 69: Could we call this from `decomp_2d_finalize`?

examples/io_test/io_bench.f90
```diff
@@ -40,11 +42,20 @@ program io_bench
...
 
    t1 = MPI_WTIME()
-   call decomp_2d_write_one(1, u1, 'io.dat')
+#ifdef ADIOS2
+   call io%open_start("./", decomp_2d_write_mode)
+#endif
+   call decomp_2d_write_one(1, u1, 'io.dat', opt_dirname="./", opt_io=io)
+#ifdef ADIOS2
+   call io%end_close
+#endif
```
pbartholomew08, line 58: Could we handle `#ifdef ADIOS2` inside the `io%open_start/close` calls?

examples/io_test/io_plane_test.f90
```diff
@@ -54,16 +55,10 @@ program io_plane_test
...
-   call decomp_2d_register_variable(io_name, "y_pencil-x_plane.dat", 2, 0, 1, mytype)
-   call decomp_2d_register_variable(io_name, "y_pencil-y_plane.dat", 2, 0, 2, mytype)
-   call decomp_2d_register_variable(io_name, "y_pencil-z_plane.dat", 2, 0, 3, mytype)
-   call decomp_2d_register_variable(io_name, "z_pencil-x_plane.dat", 3, 0, 1, mytype)
-   call decomp_2d_register_variable(io_name, "z_pencil-y_plane.dat", 3, 0, 2, mytype)
-   call decomp_2d_register_variable(io_name, "z_pencil-z_plane.dat", 3, 0, 3, mytype)
+
+   call decomp_2d_io_register_var2d("x_pencil-x_plane.dat", 1, real_type)
+   call decomp_2d_io_register_var2d("y_pencil-y_plane.dat", 2, real_type)
+   call decomp_2d_io_register_var2d("z_pencil-z_plane.dat", 3, real_type)
```
pbartholomew08, line 61: Where do we pass information about coarsening? ADIOS2 needs this during variable registration.

examples/io_test/io_bench.f90
pbartholomew08: No, as with previous design, this allows creating a file per variable with MPIIO when not opened or appending multiple data to a single file when an open file exists. By having an `io` object though, perhaps we can improve on this design? e.g.
`
io%open_start(..., opt_write_multiple=.true.) ! Writes multiple data to one file
io%open_start(..., opt_write_multiple=.false.) ! Writes single variable to file (default?)
decomp_2d_write_one(..., opt_io=io) ! Writes according to options set on io object, opens single file if none open
`

pbartholomew08 (commented):
I think it is looking good, comments intended for discussion

examples/io_test/io_bench.f90
mathrack: As a side remark, this might introduce circular dependencies. Currently, decomp_2d does not depends on decomp_2d_io.

pbartholomew08: I'd be happy to move init and finalize as suggested, however if it introduces circular dependencies obviously this is not an option.

Also I remember some discussion about being able to use decomp2d without MPI and now you've pointed it out I think even without circular dependencies this would prevent that.

In which case I suggest we leave as is.

examples/io_test/io_bench.f90
mathrack: The current design for `decomp_2d_write_one` allows to

1. Write only one variable in one file. This is possible if a IO object is provided or if a path is provided.
2. Write several variables in one file. This is possible if a IO object is provided.

I have tried to avoid mandatory IO objects for external codes using MPI... To move the IO open / start / end / close calls inside the subroutine `decomp_2d_write_one`, I think we should have one interface for writing one variable in one file. And another interface for writing several variables in one file ?

pbartholomew08: ah, I thought we wanted to make the IO objects mandatory.

examples/io_test/io_plane_test.f90
mathrack: My plan is to allow coarsening using a dedicated decomp_info object and the interpolation module. When the variable is registered, the decomp_info object of the coarse grid can be provided.

examples/io_test/io_bench.f90
mathrack: We could, the opt_dirname argument could then be removed. And some code dedicated to cases without the IO object argument could be removed. https://github.com/2decomp-fft/2decomp-fft/blob/refactor_io/src/io.f90#L1107

mathrack: As a side remark, I think the lines
```
#ifdef ADIOS2
   call io%open_start("./", decomp_2d_write_mode)
#endif
```
and
```
#ifdef ADIOS2
   call io%end_close
#endif
```
are not mandatory for ADIOS2 builds : line 1107 in io.f90 will create a IO object if none is provided.

mathrack: Fixed in https://github.com/2decomp-fft/2decomp-fft/pull/269/commits/d35d0ecb24a5b8abbb8999c73f113960cdcc0de9

examples/io_test/io_visu.f90
```diff
@@ -137,17 +139,13 @@ subroutine write_data()
 
       ! Standard I/O pattern - file per field
 #ifdef ADIOS2
-      call decomp_2d_open_io(io_name, "out", decomp_2d_write_mode)
-      call decomp_2d_start_io(io_name, "out")
+      call io%open_start("out", decomp_2d_write_mode)
```
pbartholomew08, line 142: Thinking about this (and I realise this reflects the current design) - do you think we should make the single/multi-field mode part of the I/O variable? That way we could eliminate these ADIOS2-specific calls. The user interface would look something like

```
output_dir = "out" ! Just for clarity
call io%open(output_dir, mode=decomp_2d_singlefield)
call decomp_2d_write_one(1, u1, "u1.dat", opt_io=io) ! Writes u1 into out/u1.dat
call decomp_2d_write_one(2, u2, "u2.dat", opt_io=io) ! Writes u2 into out/u2.dat
...
call io%close()
```

and similarly for the multi-field/single file case.

Happy to try an implementation if you think it sounds good?

mathrack: As a side remark, the muli-field mode is specific to MPI. Thus, we could have `decomp_2d_io_write_one` that does the single field IO. And `decomp_2d_io_mpi_write_var` that does the multiple fields in one file.

pbartholomew08:
Just to check I've got the right idea - user code would use these file generators?

mathrack:
> Just to check I've got the right idea - user code would use these file generators?

When looking at the high-level code for the IO, one can clearly see some code is duplicated. I think the python scripts can be useful to change the high-level subroutines and automatically generate the code. However, this is not really designed for users but rather for developers of 2decomp.

When changing the high-level interface, it is a bit painful to replicate the modification 4 times. The python scripts might help.

pbartholomew08:
I think if it's passing all the tests then we go for it - the classes brings an improvement and we can still refine things later, this doesn't make it set in stone.

mathrack:
> I think if it's passing all the tests then we go for it - the classes brings an improvement and we can still refine things later, this doesn't make it set in stone.

Sounds good to me. Let's keep it draft until @rfj82982 takes a look. I think some minor cleaning is needed before switching from draft to regular PR.

mathrack:
@pbartholomew08 @rfj82982

rfj82982:
I'll test this on multi-GPU as soon as I have a bit of time and approve the PR

src/io_object.f90
```diff
@@ -0,0 +1,473 @@
...
+      else
+         ! Safety check
+         call decomp_2d_abort(__FILE__, __LINE__, family%type, &
+                              "Invalid family "//family%label)
+      end if
+
+      ! Open IO
+      if (family%type == decomp_2d_io_mpi) then
+         call MPI_FILE_OPEN(decomp_2d_comm, io_dir, &
+                            access_mode, MPI_INFO_NULL, &
```
mathrack, line 160: The external code could provide some hints to mpi_file_open here.

src/io.f90
```diff
@@ -149,1391 +231,1941 @@ subroutine decomp_2d_io_finalise()
...
+      call MPI_TYPE_SIZE(data_type, type_bytes, ierror)
+      if (ierror /= 0) call decomp_2d_abort(__FILE__, __LINE__, ierror, "MPI_TYPE_SIZE")
+
+      ! Do the MPI IO
+      call MPI_TYPE_CREATE_SUBARRAY(3, sizes, subsizes, starts, &
+                                    MPI_ORDER_FORTRAN, data_type, newtype, ierror)
+      if (ierror /= 0) call decomp_2d_abort(__FILE__, __LINE__, ierror, "MPI_TYPE_CREATE_SUBARRAY")
+      call MPI_TYPE_COMMIT(newtype, ierror)
+      if (ierror /= 0) call decomp_2d_abort(__FILE__, __LINE__, ierror, "MPI_TYPE_COMMIT")
+      call MPI_FILE_SET_VIEW(io%fh, io%disp, data_type, newtype, 'native', MPI_INFO_NULL, ierror)
```
mathrack, line 1539: The external code could provide some hints to mpi_file_set_view here.

mathrack:
> I think the PR is ready for merging. I will approve but I will not merge : I wrote most of it, somebody else should check and merge.
> 
> Some modifications of the IO we could add in this PR or in a future one :
> 
>     * The object of type `d2d_io` could be mandatory for reading / writing stuff. The interface and the code would be slightly simpler.
> 
>     * The routines `decomp_2d_write_one` and `decomp_2d_read_one`could be dedicated to writing / reading one file. Additional routines `decomp_2d_mpi_write_var` / `decomp_2d_mpi_read_var` could be used to write multiple arrays inside a given file.
> 
>     * Using the combination of the two items above, we could probably remove `#ifdef ADIOS2` code from the examples.
> 
>     * One example could be provided to illustrate the grid interpolation capability. This was designed to allow IO on grids coarser than the main one.
> 
>     * The routines `decomp_2d_write_plane` could be dedicated to writing planes. The interface and the code would be slightly simpler (the input array can be used for transpose operations). Dedicated routines `decomp_2d_write_block` could be used to write stacked planes (the input array can not be used for transpose operations) and would replace `decomp_2d_write_outflow`.
> 
>     * Routines `decomp_2d_read_plane` and `decomp_2d_read_block` could be added. The latter would replace `decomp_2d_read_inflow`.
> 
>     * Routines `decomp_2d_write_scalar` and  `decomp_2d_read_scalar` could be renamed `decomp_2d_mpi_write_scalar` and `decomp_2d_mpi_read_scalar` because they are not compatible with ADIOS2 IO.

In addition to the comments above, I think the external code could provide some hints for MPI IO using the mpi_info object. Not sure if the hints should be provided when opening the file or when calling `MPI_FILE_SET_VIEW`. This could significantly improve the IO performance on some machines. On top of that, I suspect some calls to `MPI_File_sync` might be needed somewhere.

ningli-amd:
Am I correct that MPI IO hints can only be set from function calls from applications to this library? I wonder how useful that is. It would be interesting to explore if certain parameters can be passed in at runtime, possibly through environmental variables. The most obvious use case is to set stripe size for LUSTRE file system. The library used to have some old code integrating the t3pio library. The source code that t3pio might point to the right direction.

mathrack:
> Am I correct that MPI IO hints can only be set from function calls from applications to this library? I wonder how useful that is. It would be interesting to explore if certain parameters can be passed in at runtime, possibly through environmental variables. The most obvious use case is to set stripe size for LUSTRE file system. The library used to have some old code integrating the t3pio library. The source code that t3pio might point to the right direction.

Hello, within this PR, MPI IO hints can be provided by the calling code as an integer at various locations (IO family object, IO object or subroutine argument). The default MPI IO hint is MPI_INFO_NULL. The 2decomp library or the calling code could use environment variables to set another default value for MPI IO hint. However, this would require a generic and robust module to read environment variables. Currently, the library 2decomp only has https://github.com/2decomp-fft/2decomp-fft/blob/refactor_io/src/decomp_2d_init_fin.f90#L208

I think this is slightly beyond the scope of the present PR. In addition to a dedicated module for extracting environment variables, it seems important to clarify the relative importance of the hints provided to `MPI_FILE_OPEN` and `MPI_FILE_SET_VIEW`. Is it required to provide the same hints for both ? Is it better to provide hints only when opening the file ? This will drive the set of environment variables used by the library.

Overall, before this PR, it was not possible to provide MPI IO hints. After this PR it is possible using integers. Using environment variables might be more flexible. However, some work is needed to properly integrate this functionality. This is left for a future PR.

rfj82982 (approved):
LGTM

PR #280, opened 2023-11-05
Test deferred writes

pbartholomew08:
Add a test for deferred writes based on `io_test` - uses temporary arrays to write two variables (`u`, `v`) in each dimension and uses both the file per variable and the single file, multiple variables patterns.


pbartholomew08:
@rfj82982 I'm not sure why there are so many changes - there should only be 2 changes : the new test file and updating the CMakeLists.txt to build it. I accidentally started from the `SubmissionJOSS` branch, if you want I can copy the files over and resubmit to make this cleaner

rfj82982:
It is ok. I'll merge this maybe with a squash to keep the commits compact.

rfj82982:
No I check there are a lot of stuffs that belong only to the JOSS branch that are only for the paper. If you can add just the two  new files to main that would be better.

pbartholomew08:
OK, closing and will open a new one

PR #206, opened 2023-05-27
Submodule for transpose

pbartholomew08:
Reopening the original pull request #204

PR #292, opened 2023-11-09
Tidy the transpose submodules

pbartholomew08:
Attempts to tidy the transpose submodules by removing preprocessor logic from the main `transpose` subroutines (some is still necessary to prevent unused variables/uncalled subroutines that are specific to preprocessor options). Goal is that top-level subroutines should be fairly clean and clear and over time we can eliminate the preprocessor logic.

Currently a work in progress with first attempt for `transpose_z_to_y_real` as illustration, depends on #206


rfj82982:
I'll have a look at the NVHPC build

pbartholomew08:
> I'll have a look at the NVHPC build

I think it was just a missing device attribute, hopefully fixed above.

rfj82982 (commented):
It looks ok to me. I'll try to have a look to have the same also for complex. Since with that we should be able to make the NCCL communication working also with complex and not only with real.

pbartholomew08:
> It looks ok to me. I'll try to have a look to have the same also for complex. Since with that we should be able to make the NCCL communication working also with complex and not only with real.

This was only to gather people's thoughts. I'll go through and do the complex (and the other orientations)

rfj82982:
We have something with complex but it involve a lots of copies and basically transposing two real arrays and not a single one of double the size. I had something with just a single array and was working only up until n array of `size(8)`, above U had memory leakage. Never figure out what was wrong.

pbartholomew08:
sure, I wanted to tidy things only here, not fix/optimise and then we could do that elsewhere. Hopefully by having tidier code that will be easier to figure out.

src/transpose_z_to_y.f90
```diff
@@ -31,30 +31,28 @@ module subroutine transpose_z_to_y_real_long(src, dst, decomp)
...
 
-#if defined(_GPU)
-      integer :: istat, nsize
+      real(mytype), dimension(:), pointer, contiguous :: wk1
+      real(mytype), dimension(:), pointer, contiguous :: wk2
+#ifdef _GPU
+      attributes(device) :: wk1, wk2
 #endif
-
+
```
mathrack, line 39: fprettify

src/transpose_z_to_y.f90
```diff
@@ -63,94 +61,49 @@ module subroutine transpose_z_to_y_real_long(src, dst, decomp)
...
-      if (ierror /= 0) call decomp_2d_abort(__FILE__, __LINE__, ierror, "MPI_ALLTOALLV")
-#endif
+      attributes(device) :: wk1, wk2
 #endif
 
-      ! rearrange receive buffer
-      call mem_merge_zy_real(wk2, d1, d2, d3, dst, dims(2), &
-                             decomp%y2dist, decomp)
+      call rearrange_sendbuf_real(src, wk1, decomp)
+      call define_recvbuf_real(src, decomp, wk1, wk2)
```
mathrack, line 102: The two other names are good and clear. This one is not so clear in my opinion. What about `perform_transpose_real` ? `sendbuf_to_recvbuf` ?

src/transpose_z_to_y.f90
```diff
@@ -476,4 +429,182 @@ subroutine mem_merge_zy_complex(in, n1, n2, n3, out, iproc, dist, decomp)
 
    end subroutine mem_merge_zy_complex
 
+!!! Utility subroutines
+    
+   subroutine rearrange_sendbuf_real(src, wk1, decomp)
```
mathrack, line 434: We could add some comments but otherwise the subroutine looks ok.

src/transpose_z_to_y.f90
```diff
@@ -476,4 +429,182 @@ subroutine mem_merge_zy_complex(in, n1, n2, n3, out, iproc, dist, decomp)
...
+      if (istat /= 0) call decomp_2d_abort(__FILE__, __LINE__, istat, "cudaMemcpy2D")
+#else
+      associate(foo => wk1, bar => decomp)
+      end associate
+#endif
+
+#endif
+   end subroutine rearrange_sendbuf_real
+
+   subroutine define_recvbuf_real(src, decomp, wk1, wk2)
```
mathrack, line 473: We could add some comments but otherwise the subroutine looks ok. See comment above about the name of the subroutine.

src/transpose_z_to_y.f90
```diff
@@ -476,4 +429,182 @@ subroutine mem_merge_zy_complex(in, n1, n2, n3, out, iproc, dist, decomp)
...
+      d1 = SIZE(dst, 1)
+      d2 = SIZE(dst, 2)
+      d3 = SIZE(dst, 3)
+
+      call mem_merge_zy_real(wk2, d1, d2, d3, dst, dims(2), &
+           decomp%y2dist, decomp)
+      
+   end subroutine rearrange_recvbuf_real
+
+!!!! Option-specific utility subroutines
```
mathrack, line 525: The code below is hell. However, I am not sure it can easily be simpler...

src/transpose_z_to_y.f90
```diff
@@ -476,4 +429,182 @@ subroutine mem_merge_zy_complex(in, n1, n2, n3, out, iproc, dist, decomp)
...
+   
+#ifndef EVEN
+   
+#ifndef _GPU
+
+   subroutine define_recvbuf_real_std(src, decomp, wk2)
```
mathrack, line 531: This subroutine could be always defined ? Is it important to surround it with preprocessor macros ?

src/transpose_z_to_y.f90
```diff
@@ -476,4 +429,182 @@ subroutine mem_merge_zy_complex(in, n1, n2, n3, out, iproc, dist, decomp)
...
+
+      implicit none
+
+      real(mytype), dimension(:, :, :), intent(IN) :: src
```
mathrack, line 535: I am not sure we need to keep name of the variables from the calling code. It might be easier to read with `src` / `dst` or `wk1` / `wk2`. However it could also be misleading...

src/transpose_z_to_y.f90
```diff
@@ -476,4 +429,182 @@ subroutine mem_merge_zy_complex(in, n1, n2, n3, out, iproc, dist, decomp)
...
+      call MPI_ALLTOALLV(src, decomp%z2cnts, decomp%z2disp, real_type, &
+                         wk2, decomp%y2cnts, decomp%y2disp, real_type, &
+                         DECOMP_2D_COMM_ROW, ierror)
+      if (ierror /= 0) call decomp_2d_abort(__FILE__, __LINE__, ierror, "MPI_ALLTOALLV")
+      
+   end subroutine define_recvbuf_real_std
+
+#else
+   ! _GPU is defined
+   subroutine define_recvbuf_real_gpu(decomp, wk1, wk2)
```
mathrack, line 550: This subroutine could be always defined ? Is it important to surround it with preprocessor macros ?

mathrack: The line
```attributes(device) :: wk1, wk2```
needs preprocessor macros

File: src/transpose_z_to_y.f90 [hunk 9]
```diff
@@ -476,4 +429,182 @@ subroutine mem_merge_zy_complex(in, n1, n2, n3, out, iproc, dist, decomp)
...
+
+   end subroutine define_recvbuf_real_gpu
+
+   ! END IFDEF _GPU
+#endif
+
+#else
+   ! EVEN is defined
+
+   subroutine define_recvbuf_real_even(src, decomp, wk1, wk2)
```
mathrack, line 584: This subroutine could be always defined ? Is it important to surround it with preprocessor macros ?

src/transpose_z_to_y.f90
mathrack: In my opinion it would be easier to read without the preprocessor macros.

mathrack (approved):

src/transpose_z_to_y.f90
pbartholomew08: Agreed on both accounts. The aim is to move the hell as far away as possible from the high-level parts. And over time we can try to remove the macros - see https://github.com/orgs/2decomp-fft/projects/2

src/transpose_z_to_y.f90
pbartholomew08: Agreed, I was just going on how the comments described this step

src/transpose_z_to_y.f90
pbartholomew08: Yes - it's unused if `_GPU` is defined, or `EVEN` is defined and the compiler will complain if it doesn't get used.

src/transpose_z_to_y.f90
pbartholomew08: Yes - it is unused if `_GPU` is not defined and the compiler will complain

src/transpose_z_to_y.f90
pbartholomew08: Yes - it is unused if `EVEN` is not defined and the compiler will complain

src/transpose_z_to_y.f90
pbartholomew08: Done

pbartholomew08:
I think this has gone down the wrong route, maybe worth revisiting after we've got rid of some of the `#ifdef`s

PR #294, opened 2023-11-10
Create CHANGELOG.md

pbartholomew08:


pbartholomew08:
I think it would be painful to retroactively do that. I've said that for v2.0 you should refer to the JOSS paper (we can update this once published) - if you want I can pull out the changes in the statement of need here?

I think we should keep it updated as we continue work

rfj82982:
Fine my me just to start now. Just a minor point we should got for the 3 version since we have it in our constant module ;-)

pbartholomew08:
is the paper not version 2?

rfj82982:
V2.0.0 to become V2.0.1 at the end as was noted that we were inconsistent between the tag, the version on the source and the paper ;))

pbartholomew08:
ok, I'll update that. The style i've gone for follows recommendations here https://keepachangelog.com/en/1.1.0/ - current work / changes is added to the "Unreleased" section, when we decide to make a release (either v2.1 or v3.0) then that section gets renamed as appropriate and a new Unreleased is started.

rfj82982 (approved):
LGTM

mathrack:
https://github.com/2decomp-fft/2decomp-fft/blob/dev/NEWS.md

PR #301, opened 2023-11-23
229 add complex number support for adios2

pbartholomew08:


src/io.f90
```diff
@@ -18,6 +18,7 @@ module decomp_2d_io
 
    integer, parameter, public :: decomp_2d_write_mode = 1, decomp_2d_read_mode = 2, &
                                  decomp_2d_append_mode = 3
+   integer, parameter, public :: d2d_real = 1, d2d_complex = 2
```
pbartholomew08, line 21: These should probably go into the constants module

PR #302, opened 2023-11-26
Allow FFT on multiple grids

mathrack:
The external code can use the multigrid capability without the new object `decomp_2d_fft_engine` : 
```
...
! Initialise the FFT with multigrid capability
call decomp_2d_fft_set_ngrid(3)
call decomp_2d_fft_init(pencil, nx, ny, nz, 1)
call decomp_2d_fft_init(pencil, nx+2, ny, nz, 2)
call decomp_2d_fft_init(pencil, nx, ny, nz+2, 3)
...
! Use the second FFT engine
call decomp_2d_fft_use_grid(2)
call decomp_2d_fft_3d(in, out)
! Use the first FFT engine
call decomp_2d_fft_use_grid(1) ! This is equivalent to "call decomp_2d_fft_use_grid()"
call decomp_2d_fft_3d(in, out)
...
call decomp_2d_fft_finalize()
```
With the new object `decomp_2d_fft_engine` : 
```
type(decomp_2d_fft_engine), save, target :: local_engine
...
! Initialise the FFT without multigrid capability
call decomp_2d_fft_init(pencil, nx, ny, nz)
! Initialise the local FFT engine. Warning, this will replace the default FFT engine in the library.
call local_engine%init(pencil, nx+2, ny, nz)
...
! Use the local FFT engine. Warning, this will replace the default FFT engine in the library
call local_engine%fft(in, out)
! Restore the main FFT engine and use it
call decomp_2d_fft_use_grid()
call decomp_2d_fft_3d(in, out)
! Use the local FFT engine. This will replace the default FFT engine.
call local_engine%use_it()
call decomp_2d_fft_3d(in, out)
...
call local_engine%fin()
call decomp_2d_fft_finalize()
```


mathrack:
This PR is ready for review. A few examples are needed to illustrate the capability. I can work on the example(s) if the PR is ok.

pbartholomew08:
I think we should stop calling this "multigrid" - multigrid is a particular solver, I think this is running FFT on multiple grids, not running a multigrid solver?

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ _e.g._ vX.Y - YYYY-MM-DD and a new "Unreleased" section started above.
 ### Added
 
 - Add profiling to transpose subroutines when `p_row` or `p_col` is equal to one. See [PR #251](https://github.com/2decomp-fft/2decomp-fft/pull/251)
+- Add multigrid capability for the FFT (experimental support currently). Based on the work of [Lionel Gelebart](https://github.com/LionelGelebart) and coworkers, see [AMITEX_FFTP](https://amitexfftp.github.io/AMITEX/index.html) and [2decomp15_mg](https://github.com/LionelGelebart/2decomp15_mg).
```
pbartholomew08, line 15: See comment re "multigrid"

src/fft_fftw3.f90
```diff
@@ -37,6 +37,29 @@ end subroutine fftw_cleanup
 
    integer, parameter, public :: D2D_FFT_BACKEND = D2D_FFT_BACKEND_FFTW3
 
+   ! Derived type with all the quantities needed to perform FFT
+   type decomp_2d_fft_engine
+      ! Engine-specific stuff
+      type(c_ptr), private :: plan(-1:2, 3)
+      ! All the engines have this
```
pbartholomew08, line 44: Would it be better to do extended types? e.g.
```
! In FFT_common?
type decomp_2d_fft_engine
  ! All the engines have this ...
end type

! In fft_fftw3
type, extends(decomp_2d_fft_engine) :: decomp_2d_fftw3_engine
  ! Engine-specific stuff
  type(c_ptr), private :: plan(-1:2, 3)
end type
```
this should cut down on duplication.

src/fft_fftw3.f90
```diff
@@ -253,13 +276,17 @@ subroutine c2r_1m_z_plan(plan1, decomp_sp, decomp_ph)
...
+   !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
    !  This routine performs one-time initialisations for the FFT engine
-!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
-   subroutine init_fft_engine
+   !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+   subroutine init_fft_engine(engine)
 
       implicit none
 
+      type(decomp_2d_fft_engine), target, intent(inout) :: engine
```
pbartholomew08, line 286: See above wrt a base type that is extended by different backends. We might then need each backend to have a "generic" subroutine that calls its specific one
```
subroutine init_fft_engine(engine)
  implicit none
  class(decomp_2d_fft_engine), target, intent(inout) :: engine ! This is generic, can be called by anyone

  call subroutine init_fftw3_engine(engine)
end subroutine


subroutine init_fftw3_engine(engine)
  implicit none
  type(decomp_2d_fftw3_engine), target, intent(inout) :: engine ! This is specific to the fftw3 type
  ...
end subroutine

File: src/fft_fftw3_f03.f90 [hunk 4]
```diff
@@ -291,6 +396,148 @@ function decomp_2d_fft_get_sp()
...
+   ! The external code can check if the engine is physical_in_x or physical_in_z
+   function decomp_2d_fft_get_format()
+
+      implicit none
+
+      integer :: decomp_2d_fft_get_format
+
+      decomp_2d_fft_get_format = format
+
+   end function decomp_2d_fft_get_format
```
pbartholomew08, line 539: I think a lot of this could be in a common file - it doesn't look specific to the FFTW3 backend?

src/fft_common.f90
```diff
@@ -2,39 +2,45 @@
...
+integer, pointer, save :: nx_fft => null(), ny_fft => null(), nz_fft => null()
 
 ! 2D processor grid
 ! FIXME this is already available in the module decomp_2d
 integer, save, dimension(2) :: dims
 
 ! Decomposition objects
 TYPE(DECOMP_INFO), pointer, save :: ph => null()  ! physical space
-TYPE(DECOMP_INFO), target, save :: sp  ! spectral space
+TYPE(DECOMP_INFO), pointer, save :: sp => null()  ! spectral space
```
pbartholomew08, line 16: was the `target` attribute wrong before?

mathrack: No

src/fft_fftw3.f90
mathrack: I am not sure about the exact syntax for extended types. It is not widely used in the code currently. The code in this PR is similar to the code in the library, and code duplication is avoided using `#include`.

src/fft_fftw3.f90
mathrack: I am not sure about the exact syntax for extended types. It is not widely used in the code currently. This conversation is similar to https://github.com/2decomp-fft/2decomp-fft/pull/302#discussion_r1415630725. I will resolve this one and leave the one above open.

src/fft_common.f90
```diff
@@ -75,93 +88,142 @@ subroutine fft_init_arg(pencil)     ! allow to handle Z-pencil input
 end subroutine fft_init_arg
 
 ! Initialise the FFT library to perform arbitrary size transforms
-subroutine fft_init_general(pencil, nx, ny, nz)
+subroutine fft_init_general(pencil, nx, ny, nz, opt_inplace, opt_inplace_r2c, opt_inplace_c2r)
```
pbartholomew08, line 91: Would `init_one` or `init_single` now be a better name?

src/fft_common.f90
```diff
@@ -75,93 +88,142 @@ subroutine fft_init_arg(pencil)     ! allow to handle Z-pencil input
...
    if (decomp_profiler_fft) call decomp_profiler_start("fft_init")
-#endif
 
    ! Safety checks
-   if (initialised) then
-      errorcode = 4
-      call decomp_2d_abort(__FILE__, __LINE__, errorcode, &
+   if (engine%initialised) then
+      call decomp_2d_abort(__FILE__, __LINE__, 4, &
                            'FFT library should only be initialised once')
```
pbartholomew08, line 138: Replace error message -> "FFT engine should only be initialised once", could also add the engine ID if possible.

src/fft_mkl.f90
```diff
@@ -422,73 +500,70 @@ subroutine fft_3d_c2c(in, out, isign)
...
-            !          else if (isign==DECOMP_2D_FFT_BACKWARD) then
-            !             status = DftiComputeBackward(c2c_y, wk2(:,1,k), wk2b(:,1,k))
-            !          end if
-            status = wrapper_c2c(c2c_y, wk2_c2c(1, 1, k), wk2b(1, 1, k), isign)
-            if (status /= 0) call decomp_2d_abort(__FILE__, __LINE__, status, "wrapper_c2c")
-         end do
-#endif
+         if (inplace) then
+            do k = 1, ph%xsz(3)
+               status = wrapper_c2c_inplace(c2c_y, wk2_c2c(:, :, k), isign)
```
pbartholomew08, line 521: Are these changes correcting errors? The `wrapper_c2c_inplace` doesn't seem to have changed?

src/fft_cufft.f90
```diff
@@ -373,21 +401,44 @@ end subroutine init_fft_engine
...
+            end do
          end do
-      end do
+
+      else
+
+         nullify (plan)
+         nullify (cufft_workspace)
+
+      end if
```
pbartholomew08, line 426: Should there be a check on if `ngrid > 1` in which case this should loop over all engines (I think)?

src/fft_fftw3.f90
```diff
@@ -301,33 +329,52 @@ subroutine init_fft_engine
...
+            end do
          end do
-      end do
+      else
 
-      call fftw_cleanup()
+         ! Clean engine-specific stuff in the module
+         nullify (plan)
+         call fftw_cleanup()
+
```
pbartholomew08, line 362: Similar comment re clean all engines if `ngrid > 1`

pbartholomew08 (commented):
I think it looks good, my only issue is calling it "multigrid" - I've opened a PR for this, happy to change the name to something else, but multigrid is a specific type of solver which this is not.

It looks like the "multigrid" is only implemented for the FFTW_F03 interface - do we check and abort this if user tries to use with another backend?

mathrack:
> I think it looks good, my only issue is calling it "multigrid" - I've opened a PR for this, happy to change the name to something else, but multigrid is a specific type of solver which this is not.
> 
> It looks like the "multigrid" is only implemented for the FFTW_F03 interface - do we check and abort this if user tries to use with another backend?

Thank you. I have merged the PR.

The PR will allow all FFT backends to run with multiple grids. The fftw_f03 backend now offers the possibility for in-place r2c and in-place c2r transforms. The changelog has been updated to clarify this (https://github.com/2decomp-fft/2decomp-fft/pull/302/commits/719ca547c80bb8326bbfbb1a8d0df43bc24d250e).

src/fft_common.f90
mathrack: Fixed in https://github.com/2decomp-fft/2decomp-fft/pull/302/commits/7033c26a9c4b2da305a0d16b34088a931f8bb1b7

src/fft_common.f90
mathrack: Fixed in https://github.com/2decomp-fft/2decomp-fft/pull/302/commits/c22ce37eb50ee668b668c70c6194b4019856a5c1.

Currently, the engine ID is not defined inside the object `decomp_2d_fft_engine`. The ID could be defined when the engine is part of the array `fft_engines`. However, the ID can hardly be defined when the engine is managed in the calling code.

If needed, we could add a ID field using an integer inside the object `decomp_2d_fft_engine`. The ID would be the number of the engine in the array `fft_engine`. Or it would be zero if the engine is managed by the external code. However I am not sure this is very useful.

src/fft_mkl.f90
mathrack: I think this modification comes from the removal of the `#ifdef OVERWRITE` preprocessor.

src/fft_cufft.f90
mathrack: All the engines are cleaned before cleaning the variables `plan` and `cufft_workspace`. This is performed in `decomp_2d_fft_finalize`, see fft_common.f90  : 

```
subroutine decomp_2d_fft_finalize

   implicit none

   integer :: igrid

   ! Clean each engine
   do igrid = 1, n_grid
      call fft_engines(igrid)%fin()
   end do

   ! Free memory
   n_grid = 0
   deallocate (fft_engines)

   ! Nullify all the pointers in the module
   nullify (format)
   nullify (nx_fft)
   nullify (ny_fft)
   nullify (nz_fft)
   nullify (ph)
   nullify (sp)
   nullify (wk2_c2c)
   nullify (wk2_r2c)
   nullify (wk13)
   nullify (inplace)
   nullify (inplace_r2c)
   nullify (inplace_c2r)

   ! Clean engine-specific stuff located in the module
   call finalize_fft_engine()

end subroutine decomp_2d_fft_finalize
```

rfj82982 (approved):
Now LGTM

PR #304, opened 2023-12-08
Add a profiler none dummy module allowing removal of the PROFILER preprocessor directives.

mathrack:
Most of the directives are kept to avoid conflicts with pending PR targeting the FFT backends, IO refactoring and transpose subroutines.


mathrack:
Works with gfortran. Broken with nvhpc.

pbartholomew08 (approved):

PR #310, opened 2023-12-22
Fix nccl23 dev

rfj82982:
Fix NCCL in the dev branch. Solution so far is to re-introduce two additional working arrays. When NCCL transpose will be done fully using pointers these two should be removed.


examples/fft_physical_x/fft_r2c_x.f90
```diff
@@ -141,12 +141,20 @@ program fft_r2c_x
    end do
    !$acc end loop
 
-   call MPI_ALLREDUCE(error, err_all, 1, real_type, MPI_SUM, MPI_COMM_WORLD, ierror)
-   err_all = err_all / (real(nx, mytype) * real(ny, mytype) * real(nz, mytype))
+   call MPI_ALLREDUCE(MPI_IN_PLACE, error, 1, real_type, MPI_SUM, MPI_COMM_WORLD, ierror)
+   error = error / (real(nx, mytype) * real(ny, mytype) * real(nz, mytype))
+
+   ! Abort if the error is too high
```
mathrack, line 147: This line is duplicated

src/decomp_2d.f90
```diff
@@ -117,6 +117,9 @@ module decomp_2d
    ! Shared real/complex buffers
 #if defined(_GPU)
    real(mytype), target, device, allocatable, dimension(:) :: work1, work2
+#if defined(_NCCL)
+   real(mytype), target, device, allocatable, dimension(:) :: work3, work4
```
mathrack, line 121: The work arrays `work3` and `work4` could be moved to the module `decomp_2d_nccl`.

They are used in the module `decomp_2d` only when they are allocated. A deallocate statement is thus missing.

src/decomp_2d_nccl.f90
```diff
@@ -28,8 +28,13 @@ module decomp_2d_nccl
    type(ncclComm), save, public :: nccl_comm_2decomp
    integer(kind=cuda_stream_kind), save, public :: cuda_stream_2decomp
 
+   ! Extra pointers for nccl complex transpose
+   real(mytype), dimension(:), device, public, pointer, contiguous :: work3_r_d, work4_r_d
```
mathrack, line 32: If we move `work3` and `work4` here, we could have `allocatable` instead of `pointer, contiguous`.

I think the public attribute is not needed if we move `work3` and `work4` here

mathrack (approved):

mathrack:
I have approved the PR, if it is not needed to move work3 and work4 to the decomp_2d_nccl module, then it is ok. Only one line is duplicated, I will remove it.

examples/fft_physical_x/fft_r2c_x.f90
mathrack: Fixed in https://github.com/2decomp-fft/2decomp-fft/pull/310/commits/6ce600ec82bde39b9a62ade92899ea06dfa639ce

src/decomp_2d_nccl.f90
rfj82982: yes we can do that. I just wanted to be consistent with the rest.

src/decomp_2d.f90
rfj82982: Yes i think i forgot to deallocate. I'll try to do it now

examples/fft_physical_x/fft_c2c_x.f90
```diff
@@ -143,11 +143,18 @@ program fft_c2c_x
...
    end do
    !$acc end loop
-   call MPI_ALLREDUCE(error, err_all, 1, real_type, MPI_SUM, MPI_COMM_WORLD, ierror)
-   err_all = err_all / (real(nx, mytype) * real(ny, mytype) * real(nz, mytype))
+   call MPI_ALLREDUCE(MPI_IN_PLACE, error, 1, real_type, MPI_SUM, MPI_COMM_WORLD, ierror)
+   error = error / (real(nx, mytype) * real(ny, mytype) * real(nz, mytype))
+
+   ! Abort if the error is too high
+   ! A large enough value is needed for the generic backend
+   if (error > epsilon(error) * 50 * ntest) then
```
pbartholomew08, line 151: Does error accumulate across all tests?

src/decomp_2d_nccl.f90
mathrack: If possible, I think it is best to define variables in the module where they are needed.

pbartholomew08: I think it's best to move them here:
1) They're specific to the NCCL backend so it's neater
2) They're intended to be temporary until we can fix the NCCL issue - prevents people from using them for something else (which will then be hard to untangle).

examples/fft_physical_x/fft_c2c_x.f90
rfj82982: No i could just multiply by 1000 and will be fine. Issue is mainly generic FFT wehre error is e-14 instead of e-16

PR #311, opened 2023-12-27
Split the full CI

mathrack:
Run a dedicated CI for each compiler. This should save time : most of the time, only one compiler / build is broken and the others are ok.


rfj82982:
I'll add also the test in single precision and after we are ok to merge.

rfj82982:
Converted to draft just to test the Runner only on the fast build. Once this is done i'll put the PR back as open

rfj82982 (approved):
LGTM

PR #318, opened 2024-01-08
Minor update in the distribute subroutine

mathrack:
The subroutine is working as expected and the last two lines are not needed.


rfj82982 (approved):
LGTM

PR #321, opened 2024-03-11
Update to be compatible with Incompact3D V5.0

rfj82982:
Updated to the main to be compatible with Incompac3d main.
Main is the default branch that Incompact3d will use.

### DIFF
--- a/.github/workflows/RunnerFastCI.yml
+++ b/.github/workflows/RunnerFastCI.yml
@@ -28,8 +28,7 @@ jobs:
       run: |
         ls -ltr ~/.bashrc
         source ~/.bashrc
-        source ~/.profile_github
-        module load CMake 
+        source ~/.profile_aliases
         module load nvhpc
         export FC=mpif90
         cmake -S . -B build -DBUILD_TESTING=ON
@@ -47,8 +46,7 @@ jobs:
       run: |
         ls -ltr ~/.bashrc
         source ~/.bashrc
-        source ~/.profile_github
-        module load CMake 
+        source ~/.profile_aliases
         module load nvhpc
         export FC=mpif90
         cmake -S . -B build -DBUILD_TESTING=ON -DBUILD_TARGET=gpu -DENABLE_NCCL=yes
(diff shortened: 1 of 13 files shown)


src/fft_common.f90
```diff
@@ -186,7 +186,7 @@ subroutine decomp_2d_fft_finalize
 
    if (nx_fft /= nx_global .or. ny_fft /= ny_global .or. nz_fft /= nz_global) then
       call decomp_info_finalize(ph)
-      deallocate (ph)
+      !deallocate (ph)
```
mathrack, line 189: Can we use `if (allocated(ph)) deallocate(ph)` instead of commenting the line ?

rfj82982: I'll try again but, if i remember well and for reason that are a mystery to me, the `if (allocated(ph))` did not work.

pbartholomew08: if it's a pointer does it need to be `if (associated(ph))`?

mathrack:
Yes, my mistake.

Mar 11, 2024 17:56:52 Paul Bartholomew ***@***.***>:

src/fft_common.f90
rfj82982: I need to try but with `associated` the if will be `TRUE` and we'll have again the error on `deallocate`

mathrack: If using `associated` the error persists with the intel compiler, please add a commented line before the commented statement to provide the exact version of the new intel compiler causing the issue.

src/halo_common.f90
```diff
@@ -140,7 +140,7 @@
           !$acc end kernels
        else if (ipencil == 3) then
           jst = decomp%zst(2); jen = decomp%zen(2)
-          ist = decomp%xst(1); ien = decomp%xen(1)
+          ist = decomp%zst(1); ien = decomp%zen(1)
```
mathrack, line 143: Shall we add a CI test covering halo operations with the global flag on ?

mathrack: Ok, I will just add the description of the preprocessor variable in the md file then.

src/fft_common.f90
pbartholomew08: After investigation this isn't necessary - `decomp_info_finalize(ph)` frees the contents of the object, the later call to `nullify` ensures it is no longer pointed to.

.github/workflows/RunnerFastCI.yml
```diff
@@ -46,37 +42,29 @@ jobs:
...
         module load foss
         export FC=mpif90
         export CC=mpicc
         export CXX=mpicxx
-        cmake -S . -B build -DCMAKE_BUILD_TYPE=Dev -DBUILD_TESTING=ON -DIO_BACKEND=adios2 -Dadios2_DIR=~/GIT_Work/ADIOS2/build_install_gcc12_cpu/opt/lib/cmake/adios2 -DFFT_Choice=fftw_f03
-        cmake --build build -j 4
-        cmake --install build 
-        ctest --test-dir build 
-        cmake -S . -B build -DNX=128 -DNY=128 -DNX=128 -DMPIEXEC_MAX_NUMPROCS=4
+        #cmake -S . -B build -DCMAKE_BUILD_TYPE=Dev -DBUILD_TESTING=ON -DIO_BACKEND=adios2 -Dadios2_DIR=~/GIT_Work/ADIOS2/build_install_gcc12_cpu/opt/lib/cmake/adios2 -DFFT_Choice=fftw_f03
```
pbartholomew08, line 65: Should we remove this line to clean up the config or is it being kept for later CI updates?

pbartholomew08 (dismissed):
LGTM, happy to merge once CI back and passing

src/fft_common.f90
mathrack: Please clarify if this is connected with the new or old intel compiler ? Please also specify the exact version of the intel compiler connected with the modification : I think it is a bug and it will probably be fixed later. A similar modification is probably needed in `fftw_f03.f90`.

rfj82982: This has been resolved and the `deallocate(ph)` statement is no more necessary

.github/workflows/RunnerFastCI.yml
rfj82982: All commented code has been removed and all functionalities have been restored

slaizet (approved):

PR #325, opened 2024-04-03
Update github actions

mathrack:


pbartholomew08:
Looks OK. Do we know if the CI failure is due to the change from github actions v3 -> v4 or a problem with the machine?

pbartholomew08:
ah I see

pbartholomew08 (approved):

PR #334, opened 2024-04-09
Fix parallel make

mathrack:


rfj82982 (approved):
LGTM

rfj82982:
I'll merge this and after I'll have a look at the tests that are failing. If we cannot fix quickly the part with ADIOS I would suggest to remove them from the CI (or keep only 1 CI with them) to avoid all this plethora of failed tests, that in essence is just a single aspect of the library. Let me know what do you think? Maybe we can have a CI just for ADIOS with all the compilers in it

rfj82982:
Yes this is also an option

rfj82982:
Just merging the main first check that only ADIOS test are failing and after merge

mathrack:
I will open a dedicated PR to skip running the ADIOS2 CI tests. We can talk about the CI tests at the beginning of the next week

Issue #337, opened 2024-04-15
DCT-DFT-DFT or nothing-DFT-DFT


BenMql:
Greetings, 

I am developing and using [Coral](https://github.com/benmql/coral), a parallel spectral solver for partial differential equations in the plane layer geometry. Coral relies on a Chebyshev-Fourier-Fourier decomposition of the variables, which would be a great (and, I believe, reasonably easy) additional feature for `2decomp`. 

[A few years ago, I modified 2decomp v1.5  (more precisely, the ./src/fft_fftw_f03.f90 file)](https://github.com/BenMql/2decomp_emptyFourierFourier/) in order to add the flag `IN_CORE_CHEBYSHEV`. When initialised with this flag, the modified 2decomp library essentially behaves similarly to `PHYSICAL_IN_Z`, except that the transform along the contiguous dimension is skipped. Hence, in-core DCT can be computed (serially) by external routines, before calling 2decomp_fft complex to real transforms. This paves the way for DCT-DFT-DFT transforms, and resolution of PDEs in the plane layer geometry.

Currently, `Coral` can be either linked against `fftw-mpi` (for a 1D, slab domain decomposition), or against the modified 2decomp v1.5 described above. The issue is that this clunky patch is based on a deprecated version of the library, and it is rather unpractical for me to patch more recent versions as they are published. 
On the other hand, it seems that it would advantageously expand the scope of 2decomp to officially support DCT-DFT-DFT (or even empty-DFT-DFT, and let the user compute the in-core DCT himself). The [modified source file](https://github.com/BenMql/2decomp_emptyFourierFourier/blob/main/src/fft_fftw3_f03.f90) may contain all the tricks necessary for the implementation (by searching for `IN_CORE_CHEBYSHEV` and `BM patch`)

In addition, I would be happy to contribute if I can.

Best,
--Ben Miquel

mathrack:
Hello,

Thank you for your suggestion, we will look into it. The current sketch is interesting but probably not very generic.

One way to make this generic could be to add an optional argument (`logical, dimension(3)`) allowing to skip a complex-to-complex 1D transform in r2c, c2r and c2c transforms.

BenMql:
Thank you for your swift reply. Your proposed solution indeed sounds sensible to me.

rfj82982:
The best would be if you can try to merge them in the current version of 2DECOMP&FFT and to a draft PR. With this we can have a more in depth look at the code and potentially merge it. 

Stefano

PR #339, opened 2024-04-17
Skipping 1D-c2c transforms within a 3D transform, fftw_f03 backend (issue #337) 

BenMql:
As suggested by @rfj82982 , I started to implement in this draft the possibility to skip some 1d-c2c transforms.
As this is a stub, I have focussed on the `fftw_f03` backend by modifying `src/fft_fftw3_f03.f90` (and marking my edits with `!bm_issue337` suffixes).
A `logical, dimension(3), optional, intent(in) :: opt_skip_XYZ_c2c` flag can be passed when calling `decomp_2d_fft_init`.

In a way very similar to how the `inplace` optional parameter is handled within the module, this information is eventually stored in three pointers: 
```
   logical, pointer, save :: skip_x_c2c => null()  
   logical, pointer, save :: skip_y_c2c => null()  
   logical, pointer, save :: skip_z_c2c => null()
```

When comes the time of computing the transforms, the `c2c_1m_x,y,z` routines are called or not depending on these pointers. Before making modifications throughout the entire module, I focused on the `fft_3d_c2r` and `fft_3d_r2c` routines, and on the `(format == PHYSICAL_IN_Z)` case first. 

I have tested that errors are within tolerance using a modified `examples/fft_physical_z/fft_r2c_z.f90` (for a proper normalization). 

There may be a more elegant way to skip transforms (perhaps at the level of fft planning?). If you believe this implementation is worth it, I would be glad to generalise it to `fft_3d_c2c` and `PHYSICAL_IN_X`.

In a near future, I would hope to be able to provide similar modifications for the gpu backends.


mathrack (dismissed):
Hello.

The draft PR is looking good in my opinion. When I suggested this optional argument, it would be present only when calling the 3D transforms. The resulting modification would be slightly simpler. However, it also makes sense to include the optional argument only at the engine level during initialization, as you did.

I would suggest to elaborate dedicated examples in the folders physical_in_x and physical_in_z.

Maybe we want to print a warning when the user is trying to skip a 1D r2c or a 1D c2r transform ?

You can add a line for the PR in the Changelog and in the Contributors.

src/fft_fftw3_f03.f90
```diff
@@ -1400,13 +1434,15 @@ subroutine fft_3d_c2r(in_c, out_r)
...
             end if
-            call c2c_1m_y(wk2_r2c, plan(2, 2))
+            if (.not.(skip_y_c2c)) call c2c_1m_y(wk2_r2c, plan(2, 2)) !bm_issue337
          else  ! in_c==wk2_r2c if 1D decomposition
-            if (inplace) then
-               call c2c_1m_y(in_c, plan(2, 2))
-            else
-               call c2c_1m_y(wk1, plan(2, 2))
-            end if
+            if (.not.(skip_y_c2c)) then !bm_issue337
```
mathrack, line 1439: It might be easier to read and more consistent with the previous lines to have
```
            if (inplace) then
               if (.not.(skip_y_c2c)) call c2c_1m_y(in_c, plan(2, 2))
            else
               if (.not.(skip_y_c2c)) call c2c_1m_y(wk1, plan(2, 2))
            end if
```

pbartholomew08: being pedantic, wouldn't you put the skip test outside?
```
if (.not. skip_y_c2c) then
  if (inplace) then
    call ...
  else
    call ...
  end if
end if
```

mathrack:
The PR should also update https://github.com/2decomp-fft/2decomp-fft/blob/dev/src/fft_log.f90 to print in the listing the skip parameters

mathrack:
After another look at the PR, I think we should significantly change it. Instead of modifying the 3D transforms (r2c, c2r and c2c), I think it will be simpler to add only one line in the subroutines `c2c_1m_x`, `c2c_1m_y` and `c2c_1m_z` to return directly when skip is true instead of calling `fftw_execute_dft` or ``fftwf_execute_dft

src/fft_fftw3_f03.f90
mathrack: The question no longer holds if we skip inside the subroutine `c2c_1m_*`

rfj82982:
Yes potentially to skip inside `c2c_1m` would be less intrusive and the FFT algorithm will look cleaner. I would just work on a branch and try to merge directly in `dev`

src/fft_fftw3_f03.f90
```diff
@@ -1038,14 +1038,16 @@ subroutine c2c_1m_x(inout, plan1)
 
       complex(mytype), dimension(:, :, :), intent(INOUT) :: inout
       type(C_PTR) :: plan1
-
+      if (skip_x_c2c) then !bm_issue337
```
mathrack, line 1041: I think you can fit the modification in one line : 
```
      if (skip_x_c2c) return !bm_issue337

#ifdef DOUBLE_PREC
      call fftw_execute_dft(plan1, inout, inout)
#else
      call fftwf_execute_dft(plan1, inout, inout)
#endif

   end subroutine c2c_1m_x
```

BenMql:
In the last commit, the skipping occurs within the `c2c_1m_x,y,z` subroutines.
 
In addition, I have (clunkily) added warning messages at execution time in `r2c_1m` and `c2r_1m` routines. I first considered doing it during init, but it seems to me that the planner is not aware of what kind of data it will be fed in the future.  

Finally I have reverted the fft_r2c_z.f90 example to what it was. I have added examples on how to skip a transform, including one example that triggers a warning message due to an illegal attempt to skip the r2c transform.

mathrack:
Thank you. Please add the new examples.

Afterward, I will polish the PR and we will wait for a colleague to review

BenMql:
> Thank you. Please add the new examples.
> 
> Afterward, I will polish the PR and we will wait for a colleague to review

Rookie mistake, they were untracked. Added now

rfj82982:
@mathrack let me know when done and I can review the PR.

mathrack:
> @mathrack let me know when done and I can review the PR.

@rfj82982 please take a look. One example is not working, I will try to understand why.

@BenMql When the tests are working we can add the example in X and merge

mathrack:
The tests should work now

mathrack:
> I'll do a full review at some point before the end of the week. Just a silly question: do we want to keep the skip option only for fftw_f03?

I think the title / scope of the present PR is explicit : the new functionality is included only in fftw_f03. I would suggest to include the functionality later in the other backends, using a dedicated PR.

@BenMql please add one line in the Changelog and another line in the Contributors

rfj82982:
OK i know where the issue is coming from. It is the NVIDIA driver?

mathrack:
Since commit https://github.com/2decomp-fft/2decomp-fft/pull/339/commits/41f4e3ac0afdcc3df57fda3342a3855883c3ba33 all the tests are working ok

rfj82982:
The VM needed just a restart to put the GPU drivers back in order. Looks like I'll have to keep an eye on the machine and do some period restarts

mathrack:
> The PR LGTM. Do you want to add still the examples in X or this will be done in another PR?

I think the examples in X should be in this PR

@BenMql can you add them ?

mathrack (approved):

PR #343, opened 2024-04-23
Restore adios2 CI

mathrack:


mathrack:
This PR is fixing

- the CI test io_read for ADIOS2 builds
- 2decomp when using the intel compiler and ADIOS2
- the subroutine `decomp_2d_io_fin`, `adios2_finalize` was called at the beginning and should be called at the end

rfj82982 (approved):
LBTM

PR #348, opened 2024-07-16
Add support for DCT / DST in the FFT backend fftw_f03

mathrack:
This PR is a sketch for adding support for DCT / DST in the FFT backend fftw_f03.

Each direction (x, y and z) can have periodicity, DCT1, DCT2, DCT3, DCT4, DST1, DST2, DST3, or DST4.

CI tests are added to test the 9**3 = 729 combinations for both physical in x and physical in z input.

This draft PR was elaborated in collaboration with Lionel Gelebart.

PR #350, opened 2024-09-18
 Add support for DCT / DST without periodicity in the FFT backend fftw_f03

mathrack:
Same PR compared with https://github.com/2decomp-fft/2decomp-fft/pull/348 except there is no periodicity

This PR is a sketch for adding support for DCT / DST in the FFT backend fftw_f03.

Each direction (x, y and z) can have periodicity, DCT1, DCT2, DCT3, DCT4, DST1, DST2, DST3, or DST4.

CI tests are added to test the 9**3 = 729 combinations for both physical in x and physical in z input.

This draft PR was elaborated in collaboration with Lionel Gelebart.

### DIFF
--- a/examples/utilities.f90
+++ b/examples/utilities.f90
@@ -17,6 +17,7 @@ module decomp_2d_testing
       module procedure decomp_2d_testing_init_2int
       module procedure decomp_2d_testing_init_5int
       module procedure decomp_2d_testing_init_6int
+      module procedure decomp_2d_testing_init_8int
    end interface decomp_2d_testing_init
 
 contains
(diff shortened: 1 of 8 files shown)


examples/utilities.f90
```diff
@@ -17,6 +17,7 @@ module decomp_2d_testing
       module procedure decomp_2d_testing_init_2int
       module procedure decomp_2d_testing_init_5int
       module procedure decomp_2d_testing_init_6int
+      module procedure decomp_2d_testing_init_8int
```
pbartholomew08, line 20: I feel like this is a design issue. Perhaps we could have a single init subroutine that accepts an array of arguments and the calling code maps these into 2decomp values?

pbartholomew08 (commented):
Haven't yet checked the nitty-gritty of the DCT/DST

mathrack:
> Haven't yet checked the nitty-gritty of the DCT/DST - is there an ordering of changes (should #348 be merged first or does order not matter?)

The PR with periodicity is currently broken, not ready for merging and I am not sure I can fix it. Therefore I have prepared a smaller PR without periodicity. This PR is ready for review. Hopefully later I can propose a new PR on top of this one with the periodicity included.

examples/utilities.f90
mathrack: In my opinion, a proper design could be to use command-line arguments such as 
```
mpirun -n 5 test --nx 32 --ny 64 --nz 128 --prow 1 --pcol 5 ....
```
However I am not sure it can be done easily without introducing extra dependencies...

I would suggest to look at this in a dedicated issue / PR

pbartholomew08: yeah, not suggesting to fix here, it just prompted the thought

rfj82982 (approved):
LGTM. I have tested also on 8 ranks and it seems all OK

Issue #354, opened 2024-11-08
[main] - Issue when writing a 3D array in y/z pencil with reduced precision using mpiio_write_real_coarse


mathrack:
The allocate statement is valid only when the input is in x pencil. It is not valid if the input is in y or z pencil

https://github.com/2decomp-fft/2decomp-fft/blob/4ff8f79098e6028dad6efd9d994b6e100d89e02a/src/io.f90#L1266

A solution could be to replace the line with

```
         allocate (varsingle(starts(1):starts(1)+subsizes(1), &
                             starts(2):starts(2)+subsizes(2), &
                             starts(3):starts(3)+subsizes(3)))
```

rfj82982:
Can you point to a test where this create an issue. We test the library in single and generally we don't have any issue.

mathrack:
If the library is compiled with double precision and the preprocessor variable `SAVE_SINGLE` is defined, the issue can be triggered. This issue is present only in the branch main. I think it is not present in the branch dev.

mathrack:
Running the following commands : 
```
FC=mpif90 cmake -S ./ -B ./build -DCMAKE_BUILD_TYPE=dev -DBUILD_TESTING=on -DFFT_Choice=fftw_f03 -DFFTW_ROOT=~/opt/fftw3/fftw-3.3.10_bld/ -DCOMPLEX_TESTS=OFF -DDOUBLE_PRECISION=ON
FC=mpif90 cmake -S ./ -B ./build -DMPIEXEC_MAX_NUMPROCS=6 -DSINGLE_PRECISION_OUTPUT=ON
cmake --build ./build --verbose
cmake --install ./build
ctest --test-dir ./build --output-on-failure
```
Leads to
```
...
      Start 16: io_bench
16/18 Test #16: io_bench .........................   Passed    0.82 sec
      Start 17: io_visu
17/18 Test #17: io_visu ..........................   Passed    3.19 sec
      Start 18: grad3d
18/18 Test #18: grad3d ...........................   Passed    0.91 sec

83% tests passed, 3 tests failed out of 18

Total Test time (real) =  21.21 sec

The following tests FAILED:
	 11 - io_test (Failed)
	 12 - io_read (Failed)
	 13 - io_tmp_test (Failed)
Errors while running CTest
```

PR #355, opened 2024-11-08
Fix IO for reduced precision and 3D array in y / z pencil

mathrack:
Closing #354


rfj82982 (commented):
LGTM and I'll release a new version of V2.0

mathrack:
Do you need a changelog since last release ?

rfj82982:
I think we should.

rfj82982 (approved):
LGTM

PR #357, opened 2024-12-10
Extend the dummy ADIOS2 interface to include most of the objects / functions / subroutines available in the ADIOS2 module

mathrack:
The external code can avoid preprocessor directives (`#ifdef ADIOS2`) and perform IO with a simpler structure : 

```
if (decomp_2d_with_adios2) call decomp_2d_register_var("io.dat", 1, real_type)

...

if (decomp_2d_with_adios2) then
   call decomp_2d_adios_write_var(io, u1, 'io.dat')
   call io%end_close
else
   call decomp_2d_write_one(1, u1, 'io.dat', opt_dirname="./")
end if
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,7 @@ _e.g._ vX.Y - YYYY-MM-DD and a new "Unreleased" section started above.
 
 ### Changed
 
-- Major update in the IO routines. The new IO interface is not backward compatible. The examples are updated accordingly. See [PR 344](https://github.com/2decomp-fft/2decomp-fft/pull/344)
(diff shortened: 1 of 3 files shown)


pbartholomew08:
> Maybe we can print a warning or an error when the dummy IO subroutines / functions are used

I think this is the best. If we are using F2018 we can do `error stop "Some error message"`

mathrack:
I will add calls to decomp_2d_abort in the dummy module

mathrack:
I have added the error message. Sorry for the typo in the last commit. Please read "used" instead of "sued".

When merging, please squash the commits to fix the typo.

pbartholomew08 (approved):

PR #358, opened 2024-12-18
Replace ifort->ifx in CI

pbartholomew08:
Intel have deprecated and removed ifort from the oneAPI distribution.


rfj82982 (approved):
LGTM

PR #364, opened 2025-01-14
Adds openMP support for fftw_f03

Nanoseb:
Enables openMP implementation of fftw_f03. 
~~Because some new functions need to be called and are only available in a openmp supported version of fftw, a cmake parameter has been added.~~ The availability of openmp in the fftw build is now automatically detected and if so the right functions are called.


mathrack:
If OpenMP is needed, the compiler flags need to be updated ? (`-fopenmp` or `-qopenmp` for GNU or Intel for instance)

Nanoseb:
In practice yes, indeed, but that doesn't seem to be needed on my local setup for some reason. I have a look why and add that.

Nanoseb:
I believe this is ready. I have tested the integration with x3d2 and it works fine there, though let me know if I there are other tests I can run.

src/transpose_x_to_y.f90
```diff
@@ -4,6 +4,7 @@
 submodule(decomp_2d) d2d_transpose_x_to_y
 
    use decomp_2d_constants, only: mytype
+   use omp_lib
```
mathrack, line 7: I will put `#ifdef FFTW_omp` and `#endif` around the module everywhere

mathrack (changes requested):

mathrack:
I think this is ready for merging, we can wait for a review from another developer. I am not sure the statements
```
#ifdef FFTW_omp
   use omp_lib
#endif
```
are needed in the transpose subroutines if we only have openmp in the comments ?

src/CMakeLists.txt
```diff
@@ -63,6 +63,11 @@ if (MPI_FOUND)
   target_link_libraries(decomp2d PRIVATE MPI::MPI_Fortran)
 endif (MPI_FOUND)
 
+find_package(OpenMP REQUIRED)
```
mathrack, line 66: If OpenMP is not available, the code can not compile. I think REQUIRED should be removed here.

Nanoseb:
All tests now pass:
```
100% tests passed, 0 tests failed out of 29
```

Nanoseb:
From fftw documentation (https://fftw.org/fftw3_doc/Usage-of-Multi_002dthreaded-FFTW.html)
>  It returns zero if there was some error (which should not happen under normal circumstances) and a non-zero value otherwise.

mathrack (approved):

PR #371, opened 2025-02-05
Upstreaming the work from uDALES eCSE

pbartholomew08:
The main change is to add support for arbitrary depth halos and for the user to be able to supply their own arrays for halo exchange. The halo update is then split into:

update_halos (the original interface extended with additional depth)
exchange_halos (this performs the communication layer)

Beyond the halo-related subroutines the halo depth is added to the alloc routines


examples/fft_physical_x/fft_c2c_x.f90
```diff
@@ -97,8 +97,8 @@ program fft_c2c_x
    ph => decomp_2d_fft_get_ph()
    !  input is X-pencil data
    ! output is Z-pencil data
-   call alloc_x(in, ph, .true.)
-   call alloc_z(out, ph, .true.)
+   call alloc_x(in, ph, opt_global=.true.)
```
mathrack, line 100: When calling alloc subroutines, the argument `opt_global` will be mandatory after this PR ?

mathrack:
Please update the changelog before merging

examples/fft_physical_x/fft_c2c_x.f90
pbartholomew08: I'll double check, in theory it shouldn't be. This could just have been an "apply everywhere" solution.

Issue #372, opened 2025-02-24
Add a memory pool


mathrack:
To allow a more efficient usage of temporary arrays, a memory pool can be used.

See the [memory pool in x3d2](https://github.com/xcompact3d/x3d2/blob/main/src/allocator.f90) for instance.

PR #376, opened 2025-04-01
Add a memory pool

mathrack:
Closing https://github.com/2decomp-fft/2decomp-fft/issues/372

The memory pool could be used when buffers are needed in the transpose operations and in the FFT backends.

### DIFF
--- a/src/decomp_2d.f90
+++ b/src/decomp_2d.f90
@@ -5,6 +5,7 @@
 module decomp_2d
 
    use MPI
+   use iso_c_binding, only : c_size_t
    use, intrinsic :: iso_fortran_env, only: real32, real64
    use factor
    use decomp_2d_constants
@@ -461,7 +462,9 @@ contains
       ! check if additional memory is required
       if (buf_size > decomp_buf_size) then
          decomp_buf_size = buf_size
-         if (.not.use_pool) then
+         if (use_pool) then
+            call decomp_pool%new_shape(decomp_pool_default_type, shp=(/buf_size/))
+         else
             if (associated(work1_r)) nullify (work1_r)
             if (associated(work2_r)) nullify (work2_r)
             if (associated(work1_c)) nullify (work1_c)
--- a/src/decomp_2d.f90
+++ b/src/decomp_2d.f90
@@ -18,6 +18,8 @@ module decomp_2d
    use decomp_2d_nccl
 #endif
 #endif
+   use m_info
+   use m_decomp_pool
 
    implicit none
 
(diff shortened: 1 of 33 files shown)


mathrack:
I think this is ready for review

ningli-amd:
Good memory management is in general a great idea. 

Without looking into the implementation details, I just feel it a bit uncomfortable at setting `complex_pool = .true.` while initializing the whole library which does not really look very clean.

mathrack:
> Good memory management is in general a great idea.
> 
> Without looking into the implementation details, I just feel it a bit uncomfortable at setting `complex_pool = .true.` while initializing the whole library which does not really look very clean.

By default, the size of the memory pool is selected to store a real array in x, y or z.

However, if the library is used to process complex numbers on the default grid, then the size of the memory blocks is two times larger. The optional parameter `complex_pool` is used to declare this specific situation. Only a subset of the examples have been modified to include the flag `complex_pool = .true.` when initializing 2decomp.

mathrack:
I am not planning to push major updates inside this PR, please let me know if some update / clarification is needed. Some stuff I could add in a future PR to extend the scope of the memory pool : 

- use the memory pool in the FFT backends when temporary arrays are needed
- add some subroutines `decomp_pool_getx`, `decomp_pool_gety` and `decomp_pool_getz` alllowing the external code to get 3D arrays with bounds `1:xsz`, `1:ysz`, `1:zsz` or `xst:xen`, `yst:yen`, `zst:zen`. Currently, the index of the 3D arrays obtained are always starting at 1.

I did my best to write clean code with plenty of comments. If needed, I can update this PR to describe more extensively the capabilities of the code, in the comments of the code or in a markdown file ([HOWTO.md](https://github.com/2decomp-fft/2decomp-fft/blob/dev/HOWTO.md) ?).

rfj82982:
I looking at the PR. Just to be clear this should be only for CPU, isn't it?

mathrack:
> I looking at the PR. Just to be clear this should be only for CPU, isn't it?

Yes. This is not ready and not used for GPUs currently.

src/decomp_2d.f90
```diff
@@ -75,11 +80,12 @@ module decomp_2d
 #endif
 
    ! derived type to store decomposition info for a given global data size
-   TYPE, public :: DECOMP_INFO
-      ! staring/ending index and size of data held by current processor
-      integer, dimension(3) :: xst, xen, xsz  ! x-pencil
-      integer, dimension(3) :: yst, yen, ysz  ! y-pencil
-      integer, dimension(3) :: zst, zen, zsz  ! z-pencil
+   TYPE, extends(info), public :: DECOMP_INFO
```
rfj82982, line 83: Shall we remove this or we keep for reference?

src/decomp_2d.f90
```diff
@@ -75,11 +80,12 @@ module decomp_2d
...
+      ! starting/ending index and size of data held by current processor
```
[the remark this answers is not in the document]
mathrack: I think it is ok to delete. Maybe I can adapt the comment for something like `! Starting / ending index and size are defined in the parent type, see info.f90`

src/fft_cufft.f90
```diff
@@ -11,6 +11,7 @@ module decomp_2d_fft
    use iso_c_binding
    use cudafor
    use cufft
+   use m_decomp_pool
```
rfj82982, line 14: Since we do not use the memory pool for GPU yet, do we need this just for the code to compile? Does it compile without this module?

rfj82982 (approved):
LGTM just a couple of comment to check.

src/fft_cufft.f90
mathrack: I think compilation will fail because `#include "fft_common.f90"`

rfj82982: I simply commented it and i managed to compile and run. This was for a CUMPI. I'll try also with NCCL

rfj82982: Yes it seems to work. Do we keep it or not?

mathrack:
I think the compilation should fail without it. Best to keep it in my opinion

Apr 4, 2025 14:59:10 Stefano Rolfo ***@***.***>:

rfj82982:
I'll wait for all the check to finish and we can merge.

rfj82982 (approved):

PR #378, opened 2025-04-16
Fix OpenMP testing and rework the CTest accordingly

rfj82982:
In this PR the issue with OpenMP is fixed and now all the CPU tests use  by default 1 and OpenMP Threading. 
This PR is also reworking the testing using a Macro. The number of MPI ranks to use during testing is also computed depending on the available ranks, number of GPUs and is the library is supporting or not the `EVEN` option


CMakeLists.txt
```diff
@@ -37,11 +36,10 @@ option(BUILD_SHARED_LIBS "Set ON to build a shared library" OFF)
 set (IO_BACKEND "mpi" CACHE STRING "Default IO backend (mpi (default) or adios2)")
 set_property(CACHE IO_BACKEND PROPERTY STRINGS mpi adios2)
 
-set(AUTHOR "Stefano Rolfo;Charles Moulinec;Paul Bartholomew")
+set(AUTHOR "The 2DECOMP&FFT team")
 set(AUTHOR_DETAILS "stefano.rolfo@stfc.ac.uk;charles.moulinec@stfc.ac.uk;p.bartholomew@epcc.ed.ac.uk")
```
pbartholomew08, line 40: Should we change this line to point to the CONTRIBUTORS file or similar?

rfj82982: I hald modified this. I want to remove the email and just point to the dev team

CMakeLists.txt
```diff
@@ -19,15 +19,14 @@ if (BUILD_TARGET MATCHES "gpu")
...
   if (ENABLE_CUDA)
-    message(STATUS "Before enable CUDA")
     enable_language(CUDA)
-    message(STATUS "After enable CUDA")
   endif()
-
 endif(BUILD_TARGET MATCHES "gpu")
 
+include(CMakeDependentOption)
+cmake_dependent_option(ENABLE_OMP "Use OpenMP with pure MPI" ON "NOT ENABLE_CUDA" OFF)
```
mathrack, line 28: Default value for ENABLE_OMP should be off ?

mathrack (changes requested):
The new ctest framework looks good. Please clarify the default value for ENABLE_OMP and update accordingly the markdown documentation.

CMakeLists.txt
rfj82982: I put this on and removed only in case of GPU compilation. Now all compilers support OpenMP therefore should be fine to have it on. The only trick is that the num threads needs to set. For our testing it is done. We need a note that if someone is using it with another application they need to do it.

mathrack: I think the default should be OpenMP off : the library was designed without OpenMP initially. Setting the default OpenMP to off will not lead to perturbations for external codes currently using 2decomp/main and switching to 2decomp/dev.

pbartholomew08: I agree, I think we should try and keep the behaviour the same as before (pure MPI) and allow users to enable threading (MPI+X) if desired

rfj82982: Ok I'll put it to OFF by default.

examples/dtt/CMakeLists.txt
```diff
@@ -1,21 +1,12 @@
...
-add_executable(dtt_x ${files_dtt_x})
-add_executable(dtt_z ${files_dtt_z})
+file(GLOB files_dtt_z fftw_f03_dtt_z.f90)
+set(case "dtt_z")
+set(case_dir "${test_dir}/${case}")
+CreateMPITest("${case_dir}" "${case}" "${files_dtt_z}" "" "" "" "1")
 
-target_link_libraries(dtt_x PRIVATE decomp2d examples_utils)
-target_link_libraries(dtt_z PRIVATE decomp2d examples_utils)
+#include_directories(${FFTW_INCLUDE_DIRS})
```
pbartholomew08, line 11: Can we delete this line or is it required for testing with FFTW?

examples/cmake/D2D_CREATE_TEST.cmake
```diff
@@ -0,0 +1,51 @@
...
+  set(loc_files "")
+  list(APPEND loc_files ${files})
+  if (BUILD_TARGET MATCHES "gpu")
+    list(APPEND loc_files "bind.sh")
+  endif()
+  foreach(ff IN LISTS loc_files)
+    file(COPY ${ff} DESTINATION ${run_dir})
+  endforeach()
+  # Create run dir and add exe
+  file(MAKE_DIRECTORY ${run_dir})
```
pbartholomew08, line 19: This seems like it could fail, should it not be
```
# Create run dir
...
# Copy stuff into run dir
...
```
?

CMakeLists.txt
rfj82982: I just pushed the OFF by default. I checked and it can be activated also using `ccmake` without the need to have it in the configure line.

examples/dtt/CMakeLists.txt
rfj82982: Yes I just kept since I was adding the line in the macro. Just forgot to remove it.

rfj82982:
What shall we do for FFTW_F03. It seems that OpenMP is needed? I need to check exactly why it is needed, if only for linking. Let me know what do you think.

pbartholomew08:
I don't think it is needed, there's just some care required when linking and the OMP one is often linked by default. Looking at the ARCHER2 module for example:
```
> ls ${FFTW_DIR}
cmake
libfftw3.a
lifftw3f.a
libfftw3f_mpi.a
libfftw3f_omp.a
...
```
you need to *not* link the `_omp` one. I've not found a good solution to this other than using `ccmake` to edit the FFTW3 variables after initial configuration.

rfj82982:
I'll have a look tomorrow on how to do this in a clean way. Thanks

rfj82982:
On my Mac with FFTW_F03 it is OK. I need to test on the VM that we use for the CI.

rfj82982:
OMP issue should be resolved. We activated the definition `-DFFTW_omp` with FFTW_F03 without checking that OMP is activated.

examples/cmake/D2D_CREATE_TEST.cmake
```diff
@@ -0,0 +1,52 @@
...
+    target_link_libraries(${case} PRIVATE OpenMP::OpenMP_Fortran)
+  endif()
+  # Install
+  install(TARGETS ${case} DESTINATION ${run_dir})
+  # Run the test
+  if (BUILD_TARGET MATCHES "gpu")
+    add_test(NAME ${case} COMMAND ${MPIEXEC_EXECUTABLE} ${MPIEXEC_NUMPROC_FLAG} ${MPI_NUMPROCS} ./bind.sh $<TARGET_FILE:${case}> ${TEST_ARGUMENTS} WORKING_DIRECTORY ${run_dir})
+  else ()  
+	  add_test(NAME ${case} COMMAND ${MPIEXEC_EXECUTABLE} ${MPIEXEC_NUMPROC_FLAG} ${MPI_NUMPROCS} $<TARGET_FILE:${case}> ${TEST_ARGUMENTS} WORKING_DIRECTORY ${run_dir})
+    set_tests_properties(${case} PROPERTIES ENVIRONMENT "OMP_NUM_THREADS=${OMP_THREADS}")
```
mathrack, line 50: Do we set the number of threads only when OpenMP is used ? If so, we could replace the line 50 with

```
     if (OPENMP_FOUND AND ENABLE_OMP)
        set_tests_properties(${case} PROPERTIES ENVIRONMENT "OMP_NUM_THREADS=${OMP_THREADS}")
     endif(OPENMP_FOUND AND ENABLE_OMP))
```

examples/io_adios/CMakeLists.txt
```diff
@@ -1,85 +1,41 @@
...
+
 file(GLOB files_adios_test io_test.f90)
-file(GLOB files_adios_read io_read.f90)
-file(GLOB files_adios_tmp_test io_tmp_test.f90)
-file(GLOB files_adios_plane_test io_plane_test.f90)
-file(GLOB files_adios_bench io_bench.f90)
-file(GLOB files_adios_visu io_visu.f90)
+set(case "io_adios_write")
+set(case_dir "${test_dir}/io_adios_write_read")
+#set(file_list "${CMAKE_SOURCE_DIR}/examples/io_adios/adios2_config.xml" )
```
mathrack, line 10: This line can be removed

mathrack (commented):
Some minor comments to resolve. I will update the markdown documentation to describe the cmake argument `ENABLE_OMP)`.

pbartholomew08 (approved):
Looks like all issues have been addressed

rfj82982:
The issue with OMP running is now resolved. All testing is performed relatively quickly and at most 5 min. However, look at the attached picture. The run took 3 min, but the overall duration is about 6 min. I looks that it is the same for all tests. I don't understand what is now the problem. 

Screenshot 2025-04-17 at 10 13 20

mathrack:
2 comments to resolve

examples/io_adios/CMakeLists.txt
rfj82982: done

examples/cmake/D2D_CREATE_TEST.cmake
rfj82982: This will be a bit more elegant, however it is not really necessary since setting `OMP_NUM_THREADS` for an executable without OpenMP does not make any difference.

rfj82982: I have added the if conditions. Just checking that everything is still running smoothly and after I'll merge.

mathrack (approved):

PR #383, opened 2025-05-27
Add an example to show the usage of the memory pool

mathrack:
The example is excluded from the test suite for GPU builds.

Apply fprettify to the source code associated with the memory pool

The external code can select the number of blocks in the memory pool before calling decomp_2d_init

It is now possible to use temporary 3D arrays with global coordinates


rfj82982 (approved):
LGTM

rfj82982 (approved):
LGTM

PR #390, opened 2025-06-26
Remove the device attribute for halo subroutines on GPU

mathrack:
The proposed PR will remove the device attribute for the argument provided to the halo subroutine in case of GPU build. 

The array provided to the halo subroutine can now be allocated before calling the halo.

The example is updated accordingly.

The PR should not break existing codes using the halo on CPU.

The main advantage is the possibility to use 2decomp halo without using CUDA Fortran.


pbartholomew08:
I will have a look - I think it's along the direction they want, they allocate their halos externally and then we separate the "exchange" operation from the "update" interface (which to preserve behaviour performs the allocation for the user).

pbartholomew08 (changes requested):
The only issue in my opinion is that the code behaves differently for CPU and GPU, I think as a termpoaray solution this is OK. 

Starting with #404 I am working to separate the current allocating halo behaviour `update_halo` from the actual data exchange (a hypothetical `exchange_halo`). I suggest we merge this current branch after resolving minor issue and then #404 and later can build upon this.

pbartholomew08 (dismissed):

mathrack:
Back to draft, to allow major update of the halo first

mathrack:
Closing : the PR has significant conflicts and is using the previous halo interface

PR #400, opened 2025-07-04
Update mysterious workaround in prepare_buffer for cases with dedicat…

mathrack:
…ed MPI communicators


rfj82982:
Do we want to merge this?

PR #406, opened 2025-07-08
Fix  OpenMP compilation for newer Cray compilers

pbartholomew08:
New Cray compilers (CCE 16+) require explicitly linking OMP, CMake currently doesn't do this.


src/decomp_2d_constants.f90
```diff
@@ -132,8 +132,8 @@ module decomp_2d_constants
    ! Major and minor version number
    !
    integer, parameter :: D2D_MAJOR = 2
-   integer, parameter :: D2D_MINOR = 2
-   logical, parameter :: D2D_RELEASE = .false.
+   integer, parameter :: D2D_MINOR = 1
+   logical, parameter :: D2D_RELEASE = .true.
```
pbartholomew08, line 136: Are these the changes you mentioned @rfj82982? Should I rever to
```
D2D_MINOR = 2
D2D_RELEASE = .false.
```
?

rfj82982: Yes we need to revert

PR #411, opened 2025-07-25
Add optional halos to allocation subroutines

pbartholomew08:
Builds on the halo exchange interface (PR #409)

PR #414, opened 2025-08-04
Add the simplified exchange_halo interfaces

### LINKED ISSUE #415: Remove duplicated information in the halo_exchange interface
The current interface for halo_exchange is

https://github.com/2decomp-fft/2decomp-fft/blob/a9ed2af1f6e0b4df91c84654f1655e8158a5dd7a/src/halo.f90#L311

with

```
      type(halo_extents_t), intent(in) :: halo_extents
      real(mytype), dimension(halo_extents%xs:halo_extents%xe, &
                              halo_extents%ys:halo_extents%ye, &
                              halo_extents%zs:halo_extents%ze), intent(inout) :: arr
#if defined(_GPU)
      attributes(device) :: arr
#endif
      integer, intent(in) :: ipencil
      integer, intent(in) :: level
      integer, dimension(3), intent(in) :: sizes
```

However, the information provided in the argument `sizes` must be compatible with the information available in `halo_extents`, `ipencil` and `level`. Instead of using an argument, the array `sizes` can be estimated directly : 

```
sizes(1) = halo_extents%xe - halo_extents%xs + 1
sizes(2) = halo_extents%ye - halo_extents%ys + 1
sizes(3) = halo_extents%ze - halo_extents%zs + 1
if (ipencil == 1) then
   sizes(2:3) = sizes(2:3) - 2*level
else if (ipencil == 2) then
   sizes(1) = sizes(1) - 2*level
   sizes(3) = sizes(3) - 2*level
else if (ipencil == 3) then
   sizes(1:2) = sizes(1:2) - 2*level
end if
```

The `level` information could be integrated in the `halo_extents` object as well.

Overall, the current interface is not optimal. In addition to a simplification of the interface, there is another issue connected to the memory pool (https://github.com/2decomp-fft/2decomp-fft/issues/412) and there is a draft PR for the GPU interface of the halo (https://github.com/2decomp-fft/2decomp-fft/pull/390).

Codes using the halo part of 2decomp are welcome to comment this issue to help us elaborate the future interface for halo operations.

pbartholomew08:
This matches the interface used by u-DALES and allows the user to specify per-axis halo depths when performing an exchange.

The majority of this is based on Sam Owens' work https://egusphere.copernicus.org/preprints/2024/egusphere-2024-96/

This should be merged after #411 and completes the u-DALES integration.


mathrack:
Please rebase and switch to a regular PR when review is needed

examples/halo_test/exchange_test.f90
```diff
@@ -287,10 +287,8 @@ subroutine test_div_haloX()
 
       halo_extents = halo_extents_t(1, [s1, s2, s3], decomp_main, 1, global)
 
-      allocate (vh(halo_extents%xs:halo_extents%xe, &
-                   halo_extents%ys:halo_extents%ye, &
-                   halo_extents%zs:halo_extents%ze))
-      allocate (wh, mold=vh)
+      call alloc_x(vh, opt_global=global, opt_levels=[0, 1, 1])
```
mathrack: I can extend the alloc interface for a scalar opt_level. If needed ?

src/halo.f90
```diff
@@ -233,13 +238,30 @@ integer function get_pencil(sizes, decomp, opt_pencil)
...
       integer, dimension(3), intent(in) :: sizes
       type(decomp_info), intent(in) :: decomp
       integer, intent(in) :: level
       logical, intent(in) :: global
 
+      integer, dimension(3) :: levels
+
+      levels(:) = level
+      levels(ipencil) = 0
+      print *, "LL ", levels
```
mathrack: debug code here

src/halo.f90
```diff
@@ -233,13 +238,30 @@ integer function get_pencil(sizes, decomp, opt_pencil)
...
+   type(halo_extents_t) function init_halo_extents(ipencil, sizes, decomp, levels, global) result(halo_extents)
+      integer, intent(in) :: ipencil
+      integer, dimension(3), intent(in) :: sizes
+      type(decomp_info), intent(in) :: decomp
+      integer, dimension(3), intent(in) :: levels
+      logical, intent(in) :: global
+
       integer :: s1, s2, s3
 
       s1 = sizes(1)
```
mathrack, line 349: I think sizes must be equal to decomp%xsz / decomp%ysz / decomp%zsz ? The argument could be removed maybe ?

src/halo.f90
```diff
@@ -306,9 +328,54 @@ type(halo_extents_t) function init_halo_extents(ipencil, sizes, decomp, level, g
...
+        case(3)
+           levels = decomp_main%zlevel
+        case default
+           levels = 0
+           call decomp_2d_abort(__FILE__, __LINE__, 10, &
+                "Invalid pencil for halo exchange")
+        end select
+     end if
+
+     sizes = size(arr) - levels
```
mathrack: `sizes = size(arr) - 2*levels` ?

src/halo.f90
```diff
@@ -352,7 +419,43 @@ subroutine halo_exchange_real(arr, ipencil, halo_extents, level, sizes)
...
```
mathrack: `sizes = size(arr) - 2*levels` ?

examples/halo_test/exchange_test.f90
pbartholomew08: I tried something similar recently with gfortran it would always expand the scalar into an array and pad it with invalid values. If you can get it to work then certainly it would be useful

src/halo.f90
pbartholomew08: You are right - or at least the behaviour between global/not global might be inconsistent. However this follows the earlier implementation (see equivalent code in `halo_common.f90` in `main` branch)
```
if (ipencil == 1) then  ! X-pencil input
     if (global) then
        xs = decomp%xst(1)
        xe = decomp%xen(1)
        ys = decomp%xst(2) - level
        ye = decomp%xen(2) + level
        zs = decomp%xst(3) - level
        ze = decomp%xen(3) + level
     else
        xs = 1
        xe = s1
        ys = 1 - level
        ye = s2 + level
        zs = 1 - level
        ze = s3 + level
     end if
```
I'd rather not risk introducing new bugs here.

src/halo.f90
pbartholomew08: Removed

src/halo.f90
pbartholomew08: Corrected

src/halo.f90
pbartholomew08: Corrected

pbartholomew08:
I've modified the halo exchange test case so that in each pencil orientation one data exchange is performed using the long interface and one using the short interface - this confirmed the bug you spotted @mathrack so should now be addressed.

examples/halo_test/exchange_test.f90
pbartholomew08: Perhaps we could open this as a new issue if it is desired?

mathrack: See https://github.com/pbartholomew08/2decomp-fft/pull/2

src/halo.f90
mathrack: This is out of the scope of the present PR. It is discussed in https://github.com/2decomp-fft/2decomp-fft/issues/415. It will be fixed in a dedicated PR.

src/halo_exchange_z_body.f90
```diff
@@ -19,9 +19,10 @@
   else
      tag_e = coord(1) + 1
   end if
-  icount = (s2 + 2 * level) * s3
-  ilength = level
-  ijump = s1 + 2 * level
+  ! icount = (s2 + 2 * level(2)) * s3
+  icount = (s2 + 2 * levels(2)) * (s3 + 2 * levels(3))
```
pbartholomew08: Just to call out this and similar changes for the review. This is ported over from the u-DALES work on 2decomp which allows for more general halos, I need to make sure I understand it myself, however currently it is passing the tests.

mathrack: Is the new interface compatible with 1D halo exchange ? For instance, if the data is in Z pencil, is it possible to perform the halo exchange operation only in Y (case `ipencil = 3` and `levels(1) = 0`) ?

If 1D halo operations are not possible, we could add a safety check when the object halo_extents_t is created.

pbartholomew08: I'm adding a test of the exchange interface based on slab decomposition

pbartholomew08: I've begun implementing a test for a slab-based decomposition. It works, however the implementation of the exchanges currently assumes that the depth of the halos is >= 1, I think we should go with the safety check you suggest. We could also add it for non-zero depths in the pencil direction as currently no halo exchange occurs here.

pbartholomew08: Actually, u-DALES requires non-zero depth in the pencil direction.

pbartholomew08:
I think this is now ready - in addition to any comments/changes requested I would suggest we wait for u-DALES to test this branch before merging

pbartholomew08:
I think this is now ready - it passes the NVHPC tests (double), assuming the rest pass it should be good to go

mathrack (approved):

Issue #420, opened 2025-11-18
invalid memory issue in Archer2, not local machine


weiwangstfc:
**Describe the bug**
The latest version displays an invalid memory issue in PrgEnv-gnu/Archer2 when calling transpose_x_to_y.
This error does not occur in my local Linux machine and Mac OS. So far, it only pops up in Archer2.

**To Reproduce**
When changing back to an old version (CPU only), it works well.
I think the issues come from the subroutine transpose_x_to_y:

real(mytype), pointer, contiguous, dimension(:) :: work1_r, work2_r

For a user defined decomp size, for example, nx+1, ny+1, nz, (nx, ny, nz are cell numbers in x, y, z), this defination does not properly mapped to the correct memory.

rfj82982:
I'll have a look in the next couple of days on Archer2

mathrack:
Hello, thank you for reporting the issue. Machine-specific bugs are difficult to solve. Please provide the exact version of the library used. As a side remark, if you are using the memory pool : the object `decomp_pool` in the module `m_decomp_pool` has a method named `print`.

Therefore, you can check the content of the memory pool before calling the transpose subroutine. Please note that by default, only MPI rank 0 will print stuff. Each MPI rank should provide a IO unit if you want to check the content of the memory pool on each rank. See https://github.com/2decomp-fft/2decomp-fft/blob/dev/src/mem_pool.f90#L910

rfj82982:
I think i have the problem as well. This is strange since i had tests on Archer dating back to July (i.e. release of V2.1 and they were ok). 
I'll check also Cray compiler and after try to sort the issue out

rfj82982:
@weiwangstfc  let us know when you have tested this fix. We'll wait for you before merging

weiwangstfc:
> [@weiwangstfc](https://github.com/weiwangstfc) let us know when you have tested this fix. We'll wait for you before merging

@rfj82982 Same error occured when I tested it in Archer2. I used the dev version of 2decomp.

rfj82982:
@weiwangstfc the branch with the fix is this [FixDispMPIALL_dev](https://github.com/2decomp-fft/2decomp-fft/tree/FixDispMPIALL_dev). The fix is not yet in dev and we would like to check it before we merge.

weiwangstfc:
@rfj82982  This branch works perfectly for my code in Archer2 now. Thanks. Wei

rfj82982:
Perfect it seems resolved

PR #422, opened 2025-11-20
Fix disp in DECOMP buffer

rfj82982:
Fix in #420  for dev branch


mathrack:
Ready for merging when the changelog is updated

mathrack (approved):

Issue #401, opened 2025-07-04
Issue in halo operations with periodicity when p_row=1 or p_col = 1


mathrack:
In halo operations, `tag_e, tag_w, tag_n, tag_s, tag_t, tag_b` are used to identify messages coming from east / west, north / south, top / bottom.

However, in case of periodicity, when `p_row = 1`, `tag_n` and `tag_s` can have the same value : 

```
       tag_s = coord(1)
       if (coord(1) == dims(1) - 1 .AND. periodic_y) then
          tag_n = 0
       else
          tag_n = coord(1) + 1
       end if
```

When `p_row=1` with periodicity, two messages will have the same tag and the same origin. The output is not well defined.

pbartholomew08:
See fix applied to `main` in PR #410

PR #426, opened 2025-12-11
401 issue in halo operations with periodicity

pbartholomew08:
This ports the fixes for halos in periodic problems when there is one rank in the periodic direction(s) originally implemented in #401 for the `main` branch.


rfj82982 (approved):
LGTM

PR #430, opened 2025-12-15
Remove deprecated subroutines and FFT backend

mathrack:


mathrack:
Ready for review @rfj82982

rfj82982:
It seems good to me. Just a minor question: do we keep the `fftw` option and have an error message as it is now or we remove the option all together.

mathrack:
Maybe I can update the error message ?

Instead of
```
FFTW3 legacy interface is not supported
```
we can have
```
FFTW3 legacy interface is not supported. Set the CMake variable FFT_Choice to fftw_f03 to use the FFTW library.
```
or something similar ?

mathrack:
> It seems good to me. Just a minor question: do we keep the `fftw` option and have an error message as it is now or we remove the option all together.

See commit 2241fcce60e9e1bfdd16d107d6e002199fc522e4

rfj82982:
OK yes better to remove the fftw option all together. When CI finish i'll merge

rfj82982 (approved):
LGTM

PR #436, opened 2026-02-27
Add check for new CrayFTN versions

pbartholomew08:
Newer CrayFTN releases have changed the submodule naming scheme. As CMake does not yet support this, we need to tell it through our cmake scripts.


mathrack (approved):
LGTM

PR #437, opened 2026-03-03
Add an example for the interpolation module

### LINKED ISSUE #434: Coarse Grid Visualisation Interfaces Missing in Latest Version
Hi,

After updating to the latest version, all features related to coarse-grid visualisation appear to no longer work.

For example, interfaces such as
use decomp_2d, only: xszV, yszV, zszV and init_coarser_mesh_statV are no longer available.

Have these functions been permanently removed, or are they expected to be restored in a future update?

Depending on your plans, I will adjust how I use these features in my code.

Thanks,
Wei

### LINKED ISSUE #435: Add an example for the interpolation capability
The 2decomp library can perform interpolation from a 3D grid to another one. Currently, a basic interpolation capability is available in https://github.com/2decomp-fft/2decomp-fft/blob/dev/src/interp.f90

We could add an example to illustrate how it can be used.

mathrack:


weiwangstfc:
Hi,

An example demonstrating the interpolation module would be very helpful for users.

It would also be useful to clearly document the limitations, available interpolation strategies, and the supported order of accuracy within the example or accompanying documentation.

For visualisation purposes, accuracy may not be critical. However, if the module is used for mesh-to-mesh interpolation during restart, both accuracy and sensitivity to grid stretching could become important considerations. It would therefore be beneficial to highlight these aspects explicitly.

Thanks,
Wei

mathrack:
> Hi,
> 
> An example demonstrating the interpolation module would be very helpful for users.
> 
> It would also be useful to clearly document the limitations, available interpolation strategies, and the supported order of accuracy within the example or accompanying documentation.
> 
> For visualisation purposes, accuracy may not be critical. However, if the module is used for mesh-to-mesh interpolation during restart, both accuracy and sensitivity to grid stretching could become important considerations. It would therefore be beneficial to highlight these aspects explicitly.
> 
> Thanks, Wei

Hi. Thank you for the quick feedback. Have you looked at the interpolation module ? Do you think the code and the comments are clear ? Have you looked at the suggested example ? Do you think the code and the comments are clear ? Have you tried to run the example ?

I will update https://github.com/2decomp-fft/2decomp-fft.github.io/tree/main/docs/source/pages to clarify the interpolation capability (limitations, available interpolation strategy, order of accuracy). Unfortunately, it will duplicate some of the stuff that is already available in the comments.

Regarding grid stretching, we could extend the basic interpolation capability to include it. However, the concept of stretching is currently nowhere present in the 2decomp library : we only deal with distributed 3D arrays (in x, y or z) assuming a regular 2D pencil distribution. On a long term perspective, in addition to the `decomp_info` object describing the numerical grid, we could have an extra object describing the physical characteristics of the grid (cartesian, cylindrical, stretching, ...). I think stretching is out of the scope of the present PR.

mathrack:
See https://github.com/2decomp-fft/2decomp-fft.github.io/pull/9 for a clarification of the interpolation capability.

rfj82982 (approved):
LGTM

PR #440, opened 2026-04-20
Support points distribution to first or last ranks

### LINKED ISSUE #439: Support configurable remainder points distribution to first or last ranks
2DECOMP uses a fixed strategy for distributing remainder points when a global dimension is not evenly divisible by the number of ranks in that direction. In my use case, I would like to control which ranks receive the extra points. For this reason, instead of having a submodule for the CaNS solver pointing directly to 2DECOMP, I have it pointing to a fork that I maintain. This works, but it is a hacky solution and makes it less smooth to stay in sync with upstream changes.

This is the corresponding diff in our fork:

```diff
diff --git a/src/decomp_2d.f90 b/src/decomp_2d.f90
index b5bf7cc..e9f31eb 100644
--- a/src/decomp_2d.f90
+++ b/src/decomp_2d.f90
@@ -996,6 +996,8 @@ contains
       size1 = data1 / proc
       nu = data1 - size1 * proc
       nl = proc - nu
+#define NEW_DISTRIBUTION
+#if !defined(NEW_DISTRIBUTION)
       st(0) = 1
       sz(0) = size1
       en(0) = size1
@@ -1012,6 +1014,21 @@ contains
       end do
       en(proc - 1) = data1
       sz(proc - 1) = data1 - st(proc - 1) + 1
+#else
+      sz(:) = data1 / proc
+      do i = 1, mod(data1, proc)
+         sz(i - 1) = sz(i - 1) + 1
+      end do
+      do i = 1, proc
+         st(i - 1) = 1 + (i - 1) * sz(i - 1)
+         en(i - 1) = i * sz(i - 1)
+      end do
+      do i = mod(data1, proc) + 1, proc
+         st(i - 1) = st(i - 1) + mod(data1, proc)
+         en(i - 1) = en(i - 1) + mod(data1, proc)
+      end do
+#endif
+#undef NEW_DISTRIBUTION
 
       return
    end subroutine distribute
```

**Describe the solution you'd like**  
It would be very useful if the library supported selecting the 1D distribution strategy through the public API, for instance, via an optional argument in the initialization routine.

For example, one could imagine allowing users to choose between strategies such as:

- assigning the remainder to the last ranks
- assigning the remainder to the first ranks

This could be exposed through an optional argument in `decomp_2d_init`, or through a small configuration variable that can be set before initialization, similar to other library options.

**Describe alternatives you've considered**  
The main alternative I have considered is patching the `distribute` routine locally to implement the distribution pattern I want. I have already tested this approach, and it works for my case, but it is not ideal because I need to keep and maintain a fork.

If you need help, I'd be happy to contribute to this feature, e.g., through an optional argument in `decomp_2d_init`, or to test it in our use case. 

If you think this is not within your roadmap, that's also fine; then please feel free to close the issue.

mathrack:
This PR allows to select the node distribution to first / last ranks for the row and column node distribution.

### DIFF
--- a/src/env_var.f90
+++ b/src/env_var.f90
@@ -40,8 +40,8 @@ contains
                                     status=ierror, trim_name=.true.)
       if (ierror == 1) then
          output = ''
-         call decomp_2d_warning(ierror, "Environment variable " // &
-                                        trim(name) //" is not defined.")
+         call decomp_2d_warning(ierror, "Environment variable "// &
+                                trim(name)//" is not defined.")
          return
       else if (ierror /= 0) then
          output = ''
@@ -50,7 +50,7 @@ contains
       end if
 
       ! Allocate the output variable
-      allocate(character(len=max(1, len)) :: output)
+      allocate (character(len=max(1, len)) :: output)
 
       ! Get the environment variable
       call get_environment_variable(name, output, &
@@ -59,7 +59,7 @@ contains
       if (ierror /= 0) then
          output = ''
          call decomp_2d_warning(__FILE__, __LINE__, ierror, &
-                                "Error when reading " // name)
+                                "Error when reading "//name)
          return
       end if
 
@@ -85,22 +85,22 @@ contains
       ! Return if the environment variable was not available
       if (raw == '') then
          output = default
-         deallocate(raw)
+         deallocate (raw)
          return
       end if
 
       ! Format for the conversion
-      allocate(character(len = 3 + max(1,len(raw))) :: fmt)
+      allocate (character(len=3 + max(1, len(raw))) :: fmt)
       fmt(1:2) = '(i'
-      write(fmt(3:len(fmt)-1), *) max(1,len(raw))
+      write (fmt(3:len(fmt) - 1), *) max(1, len(raw))
       fmt(len(fmt):len(fmt)) = ')'
 
       ! Integer convertion
-      read(raw, fmt, iostat=ierror) output
+      read (raw, fmt, iostat=ierror) output
 
       ! Free memory
-      deallocate(fmt)
-      deallocate(raw)
+      deallocate (fmt)
+      deallocate (raw)
 
    end function get_env_var_int
 
--- a/src/env_var.f90
+++ b/src/env_var.f90
@@ -11,9 +11,6 @@ module m_env_var
    ! Default is private
    private
 
-   ! Default value if no default value is provided
-   integer, parameter :: env_var_default_int = 0
-
    public :: get_env_var
 
    interface get_env_var
--- a/src/env_var.f90
+++ b/src/env_var.f90
@@ -11,7 +11,7 @@ submodule(decomp_2d) m_env_var
 contains
 
    ! Extract the raw environment variable
-   function get_env_var_char(name) result(output)
+   module function get_env_var_char(name) result(output)
 
       implicit none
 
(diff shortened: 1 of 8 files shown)


mathrack:
If needed, I can update the PR to allow an environment variable to select the default node repartition at runtime.

rfj82982:
I had a look at the development. Just to check if i understood correctly. The only two options are to have the odd size only in the first or in the last position and use can choose between these two. Is this correct?

rfj82982:
Just to check would it be useful to have a distribution like this
```
do i = 0, mod(data1, proc) - 1 
    if (mod(i, 2) == 0) then
      idx = i / 2
    else
      idx = (proc - 1) - (i / 2)
    end if
    sz(idx) = sz(idx) + 1 
end do
```
If i have done the indexing correctly this will start to add the extra point on the two edges and progressively move towards the centre.

mathrack:
> I had a look at the development. Just to check if i understood correctly. The only two options are to have the odd size only in the first or in the last position and use can choose between these two. Is this correct?

If possible, the nodes are distributed evenly. If the number of nodes / CPUs does not allow an even repartition, we currently distribute the extra nodes as follow : 

- one extra node per CPU
- only the last CPUs of the row or column have the extra node

The PR allow the external code to replace the default repartition with : 

- one extra node per CPU
- only the first CPUs of the row or column have the extra node.

> Just to check would it be useful to have a distribution like this
> 
> ```
> do i = 0, mod(data1, proc) - 1 
>     if (mod(i, 2) == 0) then
>       idx = i / 2
>     else
>       idx = (proc - 1) - (i / 2)
>     end if
>     sz(idx) = sz(idx) + 1 
> end do
> ```
> 
> If i have done the indexing correctly this will start to add the extra point on the two edges and progressively move towards the centre.

On one hand, we can easily add a third option named `custom` to allow this kind of repartition. On the other, I am not sure it will be relevant / efficient. In addition, it is slightly beyond the scope of the associated issue.

Extra stuff I can add to the PR : 

- allow environment variables to select the node repartition at runtime
- update the examples to illustrate this new capability

mathrack:
Ready for review if CI is ok

src/env_var.f90
```diff
@@ -0,0 +1,104 @@
...
+
+      ! Local variables
+      integer :: len, ierror
+
+      ! Get the size of the environment variable
+      call get_environment_variable(name, length=len, &
+                                    status=ierror, trim_name=.true.)
+      if (ierror == 1) then
+         output = ''
+         call decomp_2d_warning(ierror, "Environment variable "// &
```
mathrack: This will print a warning each time we try to read an environment variable and it is not defined. We can

1. Keep the warning as currently defined in the code but it will be a source of noise
2. Remove the warning in this situation
3. Print the warning only if `decomp_debug` is equal or above `D2D_DEBUG_LEVEL_WARN`

Debug builds have `decomp_debug` above `D2D_DEBUG_LEVEL_WARN`. Production builds have `decomp_debug` below.

mathrack: @p-costa ?

p-costa: Hi @mathrack, I’d go for:

- 3 if these environment variables are expected to be set by the user (i.e. missing them indicates a potential misconfiguration).

- If instead they are purely optional runtime toggles for the library’s behavior, I say 2 makes most sense to avoid unnecessary noise.

rfj82982 (approved):
LGTM

PR #444, opened 2026-05-12
Added OpenMP GPU transpose path for AMD offload

lbyuan0831:


src/decomp_2d.f90
```diff
@@ -65,6 +65,17 @@ module decomp_2d
    integer(kind(D2D_DEBUG_LEVEL_OFF)), public, save :: decomp_debug = D2D_DEBUG_LEVEL_OFF
 #endif
 
+   double precision, save, public :: d2d_t_y2z_pack = 0.d0
```
mathrack, line 68: I think the best way to add timers for internal operations is to update the profilers.

The Caliper profiler can be use directly to measure the pack / unpack / mpi operations. Currently, it profiles the transpose operations.

The file `profiler_none` can be updated to allow basic monitoring.

PR #445, opened 2026-05-15
Add generic profiling

mathrack:


src/profiler_none.f90
```diff
@@ -135,10 +182,141 @@ subroutine decomp_profiler_end_char(timer_name)
...
+      ! Get the IO unit
+      if (nrank == 0) open(newunit=io_unit, file='decomp_2d_perf.log', form='formatted')
+
+      do id = 1, ncur_timers
+         ! Compute min, max and average
+         time = timer(id)
+         call decomp_2d_mpi_allreduce(time, timer_min, MPI_MIN)
+         call decomp_2d_mpi_allreduce(time, timer_max, MPI_MAX)
+         call decomp_2d_mpi_allreduce(time, timer_avg, MPI_SUM)
+         timer_avg = timer_avg / timer_n(id)
```
lbyuan0831: This does not seem to give the right value, as timer_avg is now summing over all ranks rather than over the count of calls.

mathrack: Fixed in e77aad51e24771b243b69b6c883b2a431c0b02cd

src/profiler_none.f90
```diff
@@ -135,10 +183,141 @@ subroutine decomp_profiler_end_char(timer_name)
...
+      ! Local variables
+      integer :: id
+      double precision :: deltaT
+
+      ! Get the ID of the provided timer
+      id = timer_find(timer_name)
+
+      ! Update the timer
+      deltaT = MPI_WTIME() - timer_start(id)
+      timer(id) = timer(id) + deltaT
```
lbyuan0831, line 198: I guess this computes the total wall time across multiple calls of a certain operation.

src/profiler_none.f90
```diff
@@ -135,10 +183,141 @@ subroutine decomp_profiler_end_char(timer_name)
...
+      ! Get the IO unit
+      if (nrank == 0) open (newunit=io_unit, file='decomp_2d_perf.log', form='formatted')
+
+      do id = 1, ncur_timers
+         ! Compute min, max and average
+         time = timer(id)
+         call decomp_2d_mpi_allreduce(time, timer_min, MPI_MIN)
+         call decomp_2d_mpi_allreduce(time, timer_max, MPI_MAX)
+         call decomp_2d_mpi_allreduce(time, timer_avg, MPI_SUM)
+         timer_avg = timer_avg / real(timer_n(id), real64) / real(nproc, real64)
```
lbyuan0831, line 312: To be consistent with timer_avg (average across multiple calls), I suggest also adding the following two lines after line 309:
timer_min = timer_min / real(timer_n(id), real64)
timer_max = timer_max / real(timer_n(id), real64)

src/profiler_none.f90
mathrack: Yes. I have updated the comment to clarify the meaning of the timer. See 7e4f4f1e0d2599d90bb054f69582a3234f64c0b8

src/profiler_none.f90
mathrack: Fixed in 4dd90d4e44120b5f84a10119413336402b4bdeb6

lbyuan0831 (approved):
👍

PR #447, opened 2026-05-24
Update c2c, r2c and c2r transforms in CuFFT

mathrack:
Improve performance in case of serial run or slab decomposition.

Remove debug code


mathrack:
Based on this PR, I think we could avoid to allocate the arrays `wk2_c2c` and `wk2_r2c` in case of 1D domain decomposition. For a single GPU run, the cuFFT backend no longer needs a working array.

Waiting for feedback from @rfj82982 or @lbyuan0831 to go further and optimize memory usage in the cuFFT backend.

rfj82982:
Hi 
i gave a quick look and it LGTM. I'll try to test on SCARF just to be sure. @lbyuan0831  could you quickly test on H100 just to be sure and particularly test is this help with the memory?

mathrack:
I am not 100% sure we can easily remove the array `wk2_c2c`. For instance, I have changed the line below

```
!$acc data create(wk2_c2c) present(in,out)
```
to

```
!$acc data present(in,out)
!$acc data if (dims(1)>1 .and. dims(2)>1) create(wk2_c2c)
```
and the compiler prints

```
690, Generating present(out(:,:,:),in(:,:,:))
691, Generating create(wk2_c2c(:,:,:)) [if not already present]
```

One way to workaround this potential OpenACC issue would be to reshape the c2c transform as follows : 

```
if (minval(dims) == 1) then
   ! 1D slab decomposition
   !$acc data present(in,out)
   perform c2c without wk2_c2c
   !$acc end data
else
   ! 2D pencil decomposition
   !$acc data create(wk2_c2c) present(in,out)
   perform c2c with wk2_c2c
   !$acc end data
end if
```

Something similar can be done for r2c and c2r. Pros & cons : the code is easier to read & slightly longer.

rfj82982:
Yes Bo and I look at the code and we noticed that the work arrays are always allocated even if not used.  What you suggests with the if can make the code a bit longer, but clearer. I'll try to have a look later to see if there is anything else that we can do to make it a bit more streamlined.

lbyuan0831:
I tested replacing
`!$acc data create(wk2_c2c) present(in,out)`
with
`!$acc data present(in,out) if (dims(1)>1 .and. dims(2)>1) create(wk2_c2c)`
and I can confirm through a simple profiling test that it can work to reduce memory use.
Although it still shows this:
```
    691, Generating present(in(:,:,:),out(:,:,:))
         Generating create(wk2_c2c(:,:,:)) [if not already present]
```
I guess this just indicates that the compiler is prepared to create wk2_c2c on device as `create(wk2_c2c)` appears in the code, but wk2_c2c is not really created in the runtime in the case that dim(1) or dim(2) equals to 1.

mathrack:
Sounds good. I will update the PR accordingly and convert the draft to a regular PR.

mathrack:
This PR is ready for review. I will try to check if I can run a larger case on a single H100.

Do you think we should extend the CI to properly test single GPU / Slab decomposition / Pencil decomposition ?

mathrack:
I did some tests on Jean-Zay with a single GPU. Currently, running the example `fft_c2c_x 1 1 1024 512 512 10` with a debug build leads to

```
0: Null pointer for tmp$r (/linkhome/rech/genppr01/uta27wh/2decomp-fft/src/fft_cufft.f90: 537)
```
There is a bug somewhere, I will try to fix it later.

mathrack:
The error is probably some integer overflow and not a bug. Following the present PR, it looks like running a large grid on a GPU is not memory-bound. The integer overflow could be in the 2decomp code or in the CuFFT code. The error suggests the overflow is in the CuFFT code. Some lines in the transpose code could also trigger integer overflow : 

https://github.com/2decomp-fft/2decomp-fft/blob/dev/src/transpose_x_to_y.f90#L43

mathrack:
For a double precision build, I have successfully tested

- in-place and out-of-place (`-DOVERWRITE`)
- using the test programs `c2c_x`, `r2c_x` and `r2c_z` with forward + backward transforms
- on a single GPU (512^3) and on 4 GPUs (1x4, 2x2 and 4x1 using 512^3)

The current PR is ready for review and seems to run fine. Do we want to upgrade the CI inside this PR to properly test both single / slab decomposition / pencil decomposition ?

I have opened a [dedicated issue](https://github.com/2decomp-fft/2decomp-fft/issues/449) for running large grids on a small number of GPUs. I think this is slightly beyond the scope of the present PR.

mathrack:
This PR is ready for review

PR #451, opened 2026-06-10
Add pause / resume profiling capability

mathrack:


mathrack:
I think this PR is ready for review

